PR #11, opened 2019-01-12
Added Travis-CI configuration

insomniacslk:
We don't have tests yet, but let's start by integrating Travis-CI (and start writing tests)


insomniacslk:
Yup, it's already being useful :) will fix once back home

PR #14, opened 2019-01-13
Unified config and plugin loading for v6 and v4

insomniacslk:
Unified configuration and plugin loading for DHCPv6 and DHCPv4. There is some code duplication in the plugin loading that I will remove in a subsequent commit.


coredhcp.go
```diff
@@ -29,39 +29,70 @@ type Server struct {
...
-			h6, err := plugin.Setup6(pluginConf.Args...)
-			if err != nil {
-				return nil, err
+	// Load DHCPv6 plugins
+	if conf.Server6 != nil {
+		for _, pluginConf := range conf.Server6.Plugins {
+			if plugin, ok := plugins.RegisteredPlugins[pluginConf.Name]; ok {
+				log.Printf("DHCPv6: loading plugin `%s`", pluginConf.Name)
+				if plugin.Setup6 == nil {
+					log.Warningf("DHCPv6: plugin `%s` has no setup function for DHCPv6")
```
pmazzini: This needs the plugin name as parameter.

coredhcp.go
```diff
@@ -29,39 +29,70 @@ type Server struct {
...
+		}
+	}
+	// Load DHCPv4 plugins. Yes, duplicated code for now.
+	// FIXME remove duplication
+	if conf.Server4 != nil {
+		for _, pluginConf := range conf.Server4.Plugins {
+			if plugin, ok := plugins.RegisteredPlugins[pluginConf.Name]; ok {
+				log.Printf("DHCPv4: loading plugin `%s`", pluginConf.Name)
+				if plugin.Setup4 == nil {
+					log.Warningf("DHCPv4: plugin `%s` has no setup function for DHCPv4")
```
pmazzini: This needs the plugin name as parameter.

coredhcp.go
insomniacslk: good catch, PR updated

pmazzini:
Is it ready for review or where you planning further changes?

insomniacslk:
Ready for review, I removed WIP from the title

config/config.go
```diff
@@ -90,56 +90,83 @@ func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
...
+	if exists := c.v.Get(fmt.Sprintf("server%d", ver)); exists == nil {
+		// it is valid to have no server configuration defined, and in this case
+		// no listening address and no error are returned.
+		return nil, nil
 	}
-	addr := c.v.GetString("server6.listen")
+	addr := c.v.GetString(fmt.Sprintf("server%d.listen", ver))
 	if addr == "" {
-		return ConfigErrorFromString("dhcpv6: missing `server6.listen` directive")
+		return nil, ConfigErrorFromString("dhcpv%v: missing `server%d.listen` directive", ver, ver)
```
pdf: nit: should be `"dhcpv%d`, rather than `"dhcpv%v`

config/config.go
```diff
@@ -90,56 +90,83 @@ func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
 	return plugins, nil
 }
 
-func (c *Config) parseV6Config() error {
-	if exists := c.v.Get("server6"); exists == nil {
-		// it is valid to have no DHCPv6 configuration defined, so no
-		// server and no error are returned
-		return nil
+func (c *Config) getListenAddress(v6 bool) (*net.UDPAddr, error) {
```
pdf: Is a bool the right argument type here? Bool args tend to be uninformative at the call site.  What about instead something like:

```go
type configFamily int // or some similar type name

const (
    familyV4 configFamily = 4
    familyV6 configFamily = 6
)

func (c *Config) getListenAddress(family configFamily) (*net.UDPAddr, error) {}
```

Then you call with:

```go
c.getListenAddress(familyV4)
```

Then you can then get rid of all the `ver` declarations and `if` statements at function start below, and just use the `configFamily` directly (or assert as int if necessary).

coredhcp.go
```diff
@@ -29,39 +29,70 @@ type Server struct {
...
+			} else {
+				return nil, nil, config.ConfigErrorFromString("DHCPv6: unknown plugin `%s`", pluginConf.Name)
 			}
-			loadedPlugins = append(loadedPlugins, plugin)
-			if h6 == nil {
-				return nil, config.ConfigErrorFromString("no DHCPv6 handler for plugin %s", pluginConf.Name)
+		}
+	}
+	// Load DHCPv4 plugins. Yes, duplicated code for now.
+	// FIXME remove duplication
```
pdf: Is the plan to dedup this before merge?

config/config.go
```diff
@@ -90,56 +90,83 @@ func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
...
+}
+
+func (c *Config) getPlugins(v6 bool) ([]*PluginConfig, error) {
 	pluginList := cast.ToSlice(c.v.Get("server6.plugins"))
 	if pluginList == nil {
-		return ConfigErrorFromString("dhcpv6: invalid plugins section, not a list")
+		return nil, ConfigErrorFromString("dhcpv6: invalid plugins section, not a list")
 	}
 	plugins, err := parsePlugins(pluginList)
+	if err != nil {
```
pdf: Assignment, if statement, and return is redundant, you can just `return parsePlugins(pluginList)` directly.

config/config.go
```diff
@@ -90,56 +90,83 @@ func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
...
 	}
 	ipStr, portStr, err := net.SplitHostPort(addr)
 	if err != nil {
-		return ConfigErrorFromString("dhcpv6: %v", err)
+		return nil, ConfigErrorFromString("dhcpv%d: %v", ver, err)
 	}
 	ip := net.ParseIP(ipStr)
-	if ip.To4() != nil {
-		return ConfigErrorFromString("dhcpv6: missing or invalid `listen` address")
+	if v6 && ip.To4() != nil {
```
pmazzini: Why not ip.To16() or checking its length instead of using To4()?

config/config.go
```diff
@@ -90,56 +90,83 @@ func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
...
-	}
-	// load plugins
+	return &listener, nil
+}
+
+func (c *Config) getPlugins(v6 bool) ([]*PluginConfig, error) {
 	pluginList := cast.ToSlice(c.v.Get("server6.plugins"))
 	if pluginList == nil {
-		return ConfigErrorFromString("dhcpv6: invalid plugins section, not a list")
+		return nil, ConfigErrorFromString("dhcpv6: invalid plugins section, not a list")
```
pmazzini: The comment mentions dhcpv6 which may not always be the case.

config/config.go
pdf: `ip.To4()` returns `nil` for all cases where `ip` does not represent a valid IPv4 address, but `ip.To16()` does not return `nil` in the case that it represents a parsed IPv4 address, and similarly I believe length will always be 16 bytes when parsed.  So I think the logic is correct even though it's a little confusing.

That said, I believe there is still a bug here in the case that a bogus address is provided - there needs to be a check that verifies `ip != nil` to determine this, before checking that it's either a valid IPv4 address (!v6), or not a valid IPv4 address (v6).

config/config.go
pdf: I'm not familiar with the code-base, but I also find it curious that `server6.plugins` is used here, and in the original code there is no `server4.plugins` counterpart.  Is this something that is shared between v4/v6, but is for some reason stored under server6 by default?

config/config.go
insomniacslk: I had the same doubt at first, but yeah, this would be much better, I'll change it

insomniacslk:
All of your feedback above is really good, thanks @pmazzini and @pdf - I clearly should not commit when I'm tired.
I wasn't convinced myself by a bunch of things in this PR, but I committed anyway because it is still an incremental improvement to the overall structure.

More in general, this project is getting more attention than I expected, and I should start treating it accordingly, i.e. with proper testing and design review.

I'll address your feedback in this PR first, and then continue with a more structured approach.

config/config.go
insomniacslk: it's a copypasta error, it should be protocol-specific, not "dhcpv6". This PR is changing probably too many things at the same time, because it's restructuring the config parsing code to be reusable, but also adding the v4 parsing.

insomniacslk:
Feedback addressed. The last thing to do (unless more changes are requested) is to remove the duplication

pdf:
LGTM so far.

insomniacslk:
Thanks for your feedback @pdf ! We'd be happy to have you in the loop in the future too :)

I will fix the code duplication before the final review

insomniacslk:
this will require re-approval anyway

insomniacslk:
On a second look it doesn't seem there's much that could be deduplicated there, mostly because plugin handlers and setup functions are different between DHCPv6 and v4. I tried a few options, including offloading it to a `Load` method in the `plugin.Plugin` object, but even in that case they cannot be collapsed in a single method. Even if I split `Plugin.Load` into `Plugin.Load6` and `Plugin.Load4` (which I don't like), they would be just one-liners that call `Setup6` and `Setup4`. If you have any advice I'd be happy to try something else. Meanwhile I'm removing the comment about duplication. Thanks! :)

PR #16, opened 2019-01-17
server_id, client: fix build

pmazzini:
Fix `server_id` plugin and client cmd after changes in the `iana` package of the dhcp library.


insomniacslk:
`cmds/client/main.go` also fails with "undefined: iana.HwTypeEthernet"

pmazzini:
> `cmds/client/main.go` also fails with "undefined: iana.HwTypeEthernet"

Right! Updated.

PR #18, opened 2019-01-18
plugins/server_Id: implemented DHCPv4 handler

insomniacslk:
The server_id plugin only supported DHCPv6. Added v4 too


coredhcp.go
```diff
@@ -101,27 +102,75 @@ func (s *Server) LoadPlugins(conf *config.Config) ([]*plugins.Plugin, []*plugins
...
+		resp, tmp dhcpv6.DHCPv6
+		stop      bool
+		err       error
 	)
+
+	// Create a suitable basic response packet
+	switch req.Type() {
+	case dhcpv6.MessageTypeSolicit:
+		tmp, err = dhcpv6.NewAdvertiseFromSolicit(req)
+	case dhcpv6.MessageTypeRequest, dhcpv6.MessageTypeConfirm, dhcpv6.MessageTypeRenew,
```
pmazzini: On some of these messages type we should check if the server ID from the request matches our server ID. Maybe we should add a TODO with it.

coredhcp.go
```diff
@@ -101,27 +102,75 @@ func (s *Server) LoadPlugins(conf *config.Config) ([]*plugins.Plugin, []*plugins
...
+		resp, tmp *dhcpv4.DHCPv4
+		err       error
+		stop      bool
+	)
+	if req.OpCode != dhcpv4.OpcodeBootRequest {
+		log.Print("MainHandler4: unsupported opcode %d. Only BootRequest (%d) is supported", req.OpCode, dhcpv4.OpcodeBootRequest)
+		return
+	}
+	tmp, err = dhcpv4.NewReplyFromRequest(req)
+	if err != nil {
```
pmazzini: Maybe add a TODO saying that we need to verify that the server ID of the request matches the one of this server.

coredhcp.go
```diff
@@ -101,27 +102,75 @@ func (s *Server) LoadPlugins(conf *config.Config) ([]*plugins.Plugin, []*plugins
...
-func (s *Server) MainHandler4(conn net.PacketConn, peer net.Addr, d *dhcpv4.DHCPv4) {
-	log.Print(d.Summary())
+func (s *Server) MainHandler4(conn net.PacketConn, peer net.Addr, req *dhcpv4.DHCPv4) {
+	var (
+		resp, tmp *dhcpv4.DHCPv4
+		err       error
+		stop      bool
+	)
+	if req.OpCode != dhcpv4.OpcodeBootRequest {
+		log.Print("MainHandler4: unsupported opcode %d. Only BootRequest (%d) is supported", req.OpCode, dhcpv4.OpcodeBootRequest)
```
pmazzini: Build in Go 1.11 fails with:
```
./coredhcp.go:150: Logger.Print call has possible formatting directive %d
```

Maybe we should print the string with req.OpCode.String().

coredhcp.go
insomniacslk: good point - this can be done in the `server_id` plugin. It will return `nil, true` if the server ID doesn't match

coredhcp.go
insomniacslk: the actual error is that I used `Print` instead of `Printf`

coredhcp.go
insomniacslk: I'll handle this in the `server_id` plugin

insomniacslk:
This latest commit pending merge of https://github.com/insomniacslk/dhcp/pull/232

PR #19, opened 2019-01-28
implemented live config read

### LINKED ISSUE #13: Detect config changes
Whenever the config file on disk has changed, detect it and reload.

This can be easily done with `viper`'s `OnConfigChange`, https://github.com/spf13/viper/blob/master/viper.go#L276 .

kperanovic:
Implemented live config read. ```viper.WatchConfig()``` will load the config, while config parsing is done in the ```func(ev fnotify.Event)``` function. Fixes #13.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,8 @@ import (
 	"net"
 	"strconv"
 	"strings"
+	"sync/atomic"
+	"unsafe"
 
 	"github.com/coredhcp/coredhcp/logger"
 	"github.com/fsnotify/fsnotify"
@@ -59,6 +61,7 @@ func Load() (*Config, error) {
 	if err := c.v.ReadInConfig(); err != nil {
 		return nil, err
 	}
+
 	if err := c.parseConfig(protocolV6); err != nil {
 		return nil, err
 	}
@@ -77,6 +80,7 @@ func Load() (*Config, error) {
 		if err := c.parseConfig(protocolV4); err != nil {
 			log.Error(err)
 		}
+		atomic.SwapPointer((*unsafe.Pointer)(unsafe.Pointer(&c)), unsafe.Pointer(c))
 	})
 
 	return c, nil
--- a/config/config.go
+++ b/config/config.go
@@ -80,7 +80,6 @@ func Load() (*Config, error) {
 		if err := c.parseConfig(protocolV4); err != nil {
 			log.Error(err)
 		}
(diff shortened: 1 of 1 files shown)


pmazzini:
@kperanovic Thanks for working on this!

config/config.go
```diff
@@ -67,6 +68,17 @@ func Load() (*Config, error) {
 	if c.Server6 == nil && c.Server4 == nil {
 		return nil, ConfigErrorFromString("need at least one valid config for DHCPv6 or DHCPv4")
 	}
+
+	c.v.WatchConfig()
+	c.v.OnConfigChange(func(e fsnotify.Event) {
+		if err := c.parseConfig(protocolV6); err != nil {
```
pmazzini: I wonder if this is thread safe since the config is being read from several goroutines. Maybe we should use something like https://golang.org/pkg/sync/atomic/#SwapPointer.

kperanovic: You're right this should be tested. I will test it and make adjustments

kperanovic: Testing confirms that there is a data race on config update, but only if separate gorutines are writting in the config file, that is also pointed in the issue https://github.com/spf13/viper/issues/174. Reading is not affected by that. After updating config, gorutines sync with the new config at the same time

pmazzini: Shouldn't this be done inside `parseConfig`? https://github.com/coredhcp/coredhcp/blob/master/config/config.go#L168

kperanovic: You're right. I pushed a fix and iplemented pointer swapping for both v4 and v6

config/config.go
```diff
@@ -164,10 +179,13 @@ func (c *Config) parseConfig(ver protocolVersion) error {
 		Listener: listenAddr,
 		Plugins:  plugins,
 	}
+
 	if ver == protocolV6 {
 		c.Server6 = &sc
```
pmazzini: I think the protection should happen here, when swapping the config.

config/config.go
```diff
@@ -164,10 +179,13 @@ func (c *Config) parseConfig(ver protocolVersion) error {
...
+		atomic.SwapPointer((*unsafe.Pointer)(unsafe.Pointer(&c.Server6)), unsafe.Pointer(c.Server6))
```
pmazzini: What is actually being swapped here? Isn't the new config being assigned on the previous line?

ianling: I also do not understand what is happening here, but I have tested it and it works properly.

PR #22, opened 2019-03-01
dhcp6: use changed package interface

hugelgupf:
To go with https://github.com/insomniacslk/dhcp/pull/253


pmazzini:
@hugelgupf Thanks!

PR #24, opened 2019-04-19
Config file now requires interface name

insomniacslk:
The interface name is necessary to create multicast listeners.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -35,8 +35,9 @@ func New() *Config {
 // ServerConfig holds a server configuration that is specific to either the
 // DHCPv6 server or the DHCPv4 server.
 type ServerConfig struct {
-	Listener *net.UDPAddr
-	Plugins  []*PluginConfig
+	Listener  *net.UDPAddr
+	Interface string
+	Plugins   []*PluginConfig
 }
 
 // PluginConfig holds the configuration of a plugin
@@ -70,6 +71,13 @@ func Load() (*Config, error) {
 	return c, nil
 }
 
+func protoVersionCheck(v protocolVersion) error {
+	if v != protocolV6 && v != protocolV4 {
+		return fmt.Errorf("invalid protocol version: %d", v)
+	}
+	return nil
+}
+
 func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
 	plugins := make([]*PluginConfig, 0)
 	for idx, val := range pluginList {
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -97,6 +98,18 @@ func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
 	return plugins, nil
 }
 
+func (c *Config) getInterface(ver protocolVersion) (string, error) {
+	directive := fmt.Sprintf("server%d.interface", ver)
```
mdlayher: Nit: should probably validate IPv4/6 so a random integer can't be passed here.

plugins/file/plugin.go
```diff
@@ -77,7 +77,36 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
 		return nil, false
 	}
 	log.Printf("Found IP address %s for MAC %s", ipaddr, mac)
-	// TODO add an OptIANA based on the above data
+	resp.AddOption(&dhcpv6.OptIANA{
+		// FIXME this must be unique per client address
+		IaId: [4]byte{0xaa, 0xbb, 0xcc, 0xdd},
+		T1:   0,
```
mdlayher: Nit: you can technically omit these because they're zero value.

config/config.go
insomniacslk: d'oh, I forgot to replicate that check from `getListenAddress`. I could probably define a new `ipVersion` type and define two constants

plugins/file/plugin.go
insomniacslk: right! I originally set it to 3600 because I confused it with ValidLifetime and PreferredLifetime, and didn't get rid of it when reverted

config/config.go
insomniacslk: added a `protoVersionCheck` function

plugins/file/plugin.go
insomniacslk: done

plugins/file/plugin.go
```diff
@@ -77,7 +77,34 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
+	})
+	resp.AddOption(&dhcpv6.OptDNSRecursiveNameServer{
+		NameServers: []net.IP{
+			// FIXME this must be read from the config file
+			net.ParseIP("2001:4860:4860::8888"),
+			net.ParseIP("2001:4860:4860::4444"),
+		},
+	})
+	if oro := req.GetOption(dhcpv6.OptionORO); len(oro) > 0 {
+		for _, code := range oro[0].(*dhcpv6.OptRequestedOption).RequestedOptions() {
```
mdlayher: Any chance you'd ever see multiple OROs in one request?

insomniacslk: According to RFC3315, it should not[1]
```
   Unless otherwise noted, each option may appear only in the options
   area of a DHCP message and may appear only once.  If an option does
   appear multiple times, each instance is considered separate and the
   data areas of the options MUST NOT be concatenated or otherwise
   combined.
```
ORO does not explicitly state that it can appear multiple times[2], so it's only allowed once. I think the correct behaviour is to ignore any occurrence other than the first one, in case of a malformed packet

[1] https://tools.ietf.org/html/rfc3315#section-22
[2] https://tools.ietf.org/html/rfc3315#section-22.7

mdlayher: SGTM.

PR #25, opened 2019-04-21
Improved DHCP v4 support

dlad:
- Allow running v4 server without v6
- Support v4 relay

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -124,11 +124,6 @@ func (c *Config) getListenAddress(ver protocolVersion) (*net.UDPAddr, error) {
 	if err := protoVersionCheck(ver); err != nil {
 		return nil, err
 	}
-	if exists := c.v.Get(fmt.Sprintf("server%d", ver)); exists == nil {
-		// it is valid to have no server configuration defined, and in this case
-		// no listening address and no error are returned.
-		return nil, nil
-	}
 	addr := c.v.GetString(fmt.Sprintf("server%d.listen", ver))
 	if addr == "" {
 		return nil, ConfigErrorFromString("dhcpv%d: missing `server%d.listen` directive", ver, ver)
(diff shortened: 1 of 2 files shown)


coredhcp.go
```diff
@@ -167,6 +167,11 @@ func (s *Server) MainHandler4(conn net.PacketConn, peer net.Addr, req *dhcpv4.DH
 	}
 
 	if resp != nil {
+		if !req.GatewayIPAddr.IsUnspecified() {
+			// TODO: make RFC8357 compliant
+			peer = &net.UDPAddr{IP: req.GatewayIPAddr, Port: dhcpv4.ServerPort}
```
pmazzini: This looks fine as a first step to support relay messages

config/config.go
```diff
@@ -172,6 +167,10 @@ func (c *Config) parseConfig(ver protocolVersion) error {
 	if err := protoVersionCheck(ver); err != nil {
 		return err
 	}
+	if exists := c.v.Get(fmt.Sprintf("server%d", ver)); exists == nil {
```
pmazzini: Why are you moving this? Was it returning an error before? Wasn't it returning nil in line 185 after getting a nil listenAddr?

coredhcp.go
pmazzini: This looks good as a first step to support relay messages

config/config.go
dlad: It was returning an error previously when the config file only had v4 configuration, because it would fail out before reaching line 185:

```
INFO[2019-04-22T11:45:30+01:00] Registering plugin "file"                    
INFO[2019-04-22T11:45:30+01:00] Registering plugin "server_id"               
INFO[2019-04-22T11:45:30+01:00] Loading configuration                        
FATA[2019-04-22T11:45:30+01:00] error parsing config: dhcpv6: missing `server6.interface` directive 
```

```
# config.yml
server4:
    listen: '0.0.0.0:67'
    interface: "eth0"
    plugins:
        - server_id: 192.168.1.12
```

PR #26, opened 2019-04-22
Provide default IAID

insomniacslk:
The IAID should be provided by the client. If not, use the last 4 bytes
of the MAC address

### DIFF
--- a/plugins/file/plugin.go
+++ b/plugins/file/plugin.go
@@ -78,8 +78,9 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
 	}
 	log.Printf("Found IP address %s for MAC %s", ipaddr, mac)
 	resp.AddOption(&dhcpv6.OptIANA{
-		// FIXME this must be unique per client address
-		IaId: [4]byte{0xaa, 0xbb, 0xcc, 0xdd},
+		// TODO the client should provide the IAID. Use the provided one, and
+		// fall back on the last 4 bytes of the MAC address otherwise.
+		IaId: [4]byte{mac[2], mac[3], mac[4], mac[5]},
 		Options: []dhcpv6.Option{
 			&dhcpv6.OptIAAddress{
 				IPv6Addr:          ipaddr,


plugins/file/plugin.go
```diff
@@ -78,8 +78,9 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
 	}
 	log.Printf("Found IP address %s for MAC %s", ipaddr, mac)
 	resp.AddOption(&dhcpv6.OptIANA{
-		// FIXME this must be unique per client address
-		IaId: [4]byte{0xaa, 0xbb, 0xcc, 0xdd},
+		// TODO the client should provide the IAID. Use the provided one, and
+		// fall back on the last 4 bytes of the MAC address otherwise.
+		IaId: [4]byte{mac[2], mac[3], mac[4], mac[5]},
```
pmazzini: Why don't we copy the one from the client? AFAIK the reply should contain the IANA Option only if the client sends one.

pmazzini: We don't need a fallback mechanism

pmazzini: The reply should have the same IAID, shouldn't it?

insomniacslk: That's the point for the TODO, I didn't have time today to implement that, but wanted to quickly implement a valid reply if no IAID is provided by the client

pmazzini: > wanted to quickly implement a valid reply

I think the only valid reply is copying the client IAID.

> if no IAID is provided by the client

The IAID should always be provided by the client

insomniacslk: I looked up at existing implementations and, yes, the server should just drop the request. I will implement IAID in the dhcp library's client, using `dhclient`'s approach of the last 4 bytes, https://source.isc.org/cgi-bin/gitweb.cgi?p=dhcp.git;a=blob;f=client/dhclient.c;h=f7486c6a754f741fecb2a2999d78778ab79a5970;hb=846d0ecce7480257723c86c59f653687217181bc#l2987

PR #33, opened 2019-07-15
Handle rapid-commit requests

insomniacslk:

### DIFF
--- a/coredhcp.go
+++ b/coredhcp.go
@@ -104,18 +104,23 @@ func (s *Server) LoadPlugins(conf *config.Config) ([]*plugins.Plugin, []*plugins
 // It will not reply if the resulting response is `nil`.
 func (s *Server) MainHandler6(conn net.PacketConn, peer net.Addr, req dhcpv6.DHCPv6) {
 	var (
-		resp, tmp dhcpv6.DHCPv6
-		stop      bool
-		err       error
+		resp dhcpv6.DHCPv6
+		stop bool
+		err  error
 	)
 
 	// Create a suitable basic response packet
 	switch req.Type() {
 	case dhcpv6.MessageTypeSolicit:
-		tmp, err = dhcpv6.NewAdvertiseFromSolicit(req.(*dhcpv6.Message))
+		m := req.(*dhcpv6.Message)
+		if m.GetOneOption(dhcpv6.OptionRapidCommit) != nil {
+			resp, err = dhcpv6.NewReplyFromMessage(m)
+		} else {
+			resp, err = dhcpv6.NewAdvertiseFromSolicit(m)
+		}
 	case dhcpv6.MessageTypeRequest, dhcpv6.MessageTypeConfirm, dhcpv6.MessageTypeRenew,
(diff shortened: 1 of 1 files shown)


coredhcp.go
```diff
@@ -104,20 +104,26 @@ func (s *Server) LoadPlugins(conf *config.Config) ([]*plugins.Plugin, []*plugins
...
-	switch req.Type() {
+	tmp := req
+	switch tmp.Type() {
 	case dhcpv6.MessageTypeSolicit:
-		tmp, err = dhcpv6.NewAdvertiseFromSolicit(req.(*dhcpv6.Message))
+		m := tmp.(*dhcpv6.Message)
+		if m.GetOneOption(dhcpv6.OptionRapidCommit) != nil {
+			tmp, err = dhcpv6.NewReplyFromMessage(m)
+		} else {
+			tmp, err = dhcpv6.NewAdvertiseFromSolicit(req.(*dhcpv6.Message))
```
pmazzini:
```
tmp, err = dhcpv6.NewAdvertiseFromSolicit(req.(*dhcpv6.Message))
```

can be changed to:

```
tmp, err = dhcpv6.NewAdvertiseFromSolicit(m)
```

It will be more consistent with what is done for the rapid commit

coredhcp.go
```diff
@@ -104,20 +104,26 @@ func (s *Server) LoadPlugins(conf *config.Config) ([]*plugins.Plugin, []*plugins
...
-		stop      bool
-		err       error
+		resp dhcpv6.DHCPv6
+		stop bool
+		err  error
 	)
 
 	// Create a suitable basic response packet
-	switch req.Type() {
+	tmp := req
```
pmazzini: Is this `tmp` variable actually needed? I think that with the `req` and `resp` it should be enough.

insomniacslk: good point, it's not really needed. I was keeping it for the rainy days in case we wanted the original packet :) but until it's needed, it's not needed

coredhcp.go
insomniacslk: good catch

coredhcp.go
insomniacslk: actually, I realized why I did that: it's due to an unrelated change that I will make in another PR (since I'm backporting the changes that I made for the LinuxBoot demo), I'll probably have to reintroduce it later, but for it's good as you suggest

coredhcp.go
```diff
@@ -112,7 +112,12 @@ func (s *Server) MainHandler6(conn net.PacketConn, peer net.Addr, req dhcpv6.DHC
 	// Create a suitable basic response packet
 	switch req.Type() {
 	case dhcpv6.MessageTypeSolicit:
-		tmp, err = dhcpv6.NewAdvertiseFromSolicit(req.(*dhcpv6.Message))
+		m := req.(*dhcpv6.Message)
+		if m.GetOneOption(dhcpv6.OptionRapidCommit) != nil {
+			tmp, err = dhcpv6.NewReplyFromMessage(m)
```
pmazzini: Maybe remove `tmp` from here and use directly `resp`?

PR #34, opened 2019-07-15
DHCPv6: handle relay-forw requests

insomniacslk:

### DIFF
--- a/coredhcp.go
+++ b/coredhcp.go
@@ -109,20 +109,26 @@ func (s *Server) MainHandler6(conn net.PacketConn, peer net.Addr, req dhcpv6.DHC
 		err  error
 	)
 
+	// decapsulate the relay message
+	msg, err := req.GetInnerMessage()
+	if err != nil {
+		log.Warningf("DHCPv6: cannot get inner message: %v", err)
+		return
+	}
+
 	// Create a suitable basic response packet
-	switch req.Type() {
+	switch msg.Type() {
 	case dhcpv6.MessageTypeSolicit:
-		m := req.(*dhcpv6.Message)
-		if m.GetOneOption(dhcpv6.OptionRapidCommit) != nil {
-			resp, err = dhcpv6.NewReplyFromMessage(m)
+		if msg.GetOneOption(dhcpv6.OptionRapidCommit) != nil {
+			resp, err = dhcpv6.NewReplyFromMessage(msg)
 		} else {
-			resp, err = dhcpv6.NewAdvertiseFromSolicit(m)
+			resp, err = dhcpv6.NewAdvertiseFromSolicit(msg)
 		}
 	case dhcpv6.MessageTypeRequest, dhcpv6.MessageTypeConfirm, dhcpv6.MessageTypeRenew,
 		dhcpv6.MessageTypeRebind, dhcpv6.MessageTypeRelease, dhcpv6.MessageTypeInformationRequest:
-		resp, err = dhcpv6.NewReplyFromMessage(req.(*dhcpv6.Message))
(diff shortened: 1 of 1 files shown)


coredhcp.go
```diff
@@ -104,11 +104,21 @@ func (s *Server) LoadPlugins(conf *config.Config) ([]*plugins.Plugin, []*plugins
...
+		stop              bool
+		err               error
 	)
 
+	// decapsulate the relay message
+	if req.IsRelay() {
+		originalReq = req
+		tmp, err := req.GetInnerMessage()
+		if err != nil {
+			log.Warningf("DHCPv6: cannot unpack relay message: %v", err)
```
pmazzini: Should it return here instead of continuing?

coredhcp.go
```diff
@@ -104,11 +104,21 @@ func (s *Server) LoadPlugins(conf *config.Config) ([]*plugins.Plugin, []*plugins
...
 	var (
-		resp dhcpv6.DHCPv6
-		stop bool
-		err  error
+		resp, originalReq dhcpv6.DHCPv6
+		stop              bool
+		err               error
 	)
 
+	// decapsulate the relay message
```
pmazzini: How about doing `msg, err := packet.GetInnerMessage()` here and you can avoid the `if req.IsRelay()`.

And then using the `msg` variable everywhere and you can avoid the type assertion of lines 125, 133 and 151.

coredhcp.go
insomniacslk: yes, I missed it :(

coredhcp.go
insomniacslk: thanks, I'll look into this :)

insomniacslk: that works, but I still need to save the original message for the re-encapsulation below

pmazzini: The original message is in function parameter, so it is still there.

insomniacslk: done

coredhcp.go
```diff
@@ -136,6 +141,16 @@ func (s *Server) MainHandler6(conn net.PacketConn, peer net.Addr, req dhcpv6.DHC
 		}
 	}
 
+	// if the request was relayed, re-encapsulate the response
+	if req.IsRelay() {
+		tmp, err := dhcpv6.NewRelayReplFromRelayForw(req.(*dhcpv6.RelayMessage), resp.(*dhcpv6.Message))
```
pmazzini: The 2nd type assertion is no longer needed.

coredhcp.go
```diff
@@ -109,20 +109,25 @@ func (s *Server) MainHandler6(conn net.PacketConn, peer net.Addr, req dhcpv6.DHC
 		err  error
 	)
 
+	// decapsulate the relay message
+	msg, err := req.GetInnerMessage()
+	if err != nil {
+		log.Warningf("DHCPv6: cannot get inner message: %v", err)
```
pmazzini: `return` here?

coredhcp.go
pmazzini: Is the `tmp` really needed?

insomniacslk: `resp` is still of `dhcpv6.DHCPv6` type, the assertion is still needed unfortunately :( (the compiler complains otherwise)

coredhcp.go
insomniacslk: right

coredhcp.go
insomniacslk: to avoid overwriting `resp`, yes, is that what you mean?

pmazzini: Oh right, my bad.

pmazzini: Is there any problem with overwriting it?

PR #35, opened 2019-07-15
Added FIXME entries

insomniacslk:


pmazzini:
We should add a FIXME for the server to check that the DUID of incoming requests match the one of the server.

insomniacslk:
I guess we should start doing that via issues. I'll merge this, then open issues

PR #40, opened 2019-08-12
IPv4 server plugin

borna-blazevic:
An example of the IPv4 server made in the CoreDHCP library. The example uses custom creation of UDP connection, so it can bind the connection to an interface and communicate to a client without an IP address successfully.


README.md
```diff
@@ -28,6 +28,30 @@ server6:
 
 See also [config.yml.example](cmds/coredhcp/config.yml.example).
 
+## Example configuration
```
pmazzini: This is duplicated (see line 11)

pmazzini:
Thanks for your contribution @borna-blazevic!

server/server.go
```diff
@@ -0,0 +1,131 @@
+package server
```
pmazzini: All this code should be part of the library, not CoreDHCP. Would you mind creating a PR for https://github.com/insomniacslk/dhcp to modify https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/server4/server.go?

It will make ti more consistent with the v6 one: https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/server6/server.go

pmazzini:
Hey @borna-blazevic, why are you closing this? I merged https://github.com/insomniacslk/dhcp/pull/310.

You can redo this on top of the library changes.

borna-blazevic:
> Hey @borna-blazevic, why are you closing this? I merged [insomniacslk/dhcp#310](https://github.com/insomniacslk/dhcp/pull/310).
> 
> You can redo this on top of the library changes.

I will redo I it, I was planning to open a new pull request, but ok, I will push here when I redo it.

pmazzini:
> I will redo I it, I was planning to open a new pull request, but ok, I will push here when I redo it.

Oh, ok, whatever fits you best. Thanks

cmds/coredhcp/main.go
```diff
@@ -6,6 +6,8 @@ import (
 	"github.com/coredhcp/coredhcp"
 	"github.com/coredhcp/coredhcp/config"
 	"github.com/coredhcp/coredhcp/logger"
+	_ "github.com/coredhcp/coredhcp/plugins/IPv4"
+	_ "github.com/coredhcp/coredhcp/plugins/example"
```
pmazzini: Is this line needed?

plugins/IPv4/plugin.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	return Handler6, Handler4, nil
+}
+func createIP(serverIP net.IP, netmask net.IP) (Record, bool) {
+
+	rand.Seed(time.Now().Unix())
+	ipserver := serverIP.To4()
+	mask := netmask.To4()
+	ip := net.IPv4(random(1, 254), random(1, 254), random(1, 254), random(1, 254)).To4()
```
pmazzini: Is this allocating a random IP? Does this fits your needs? What is your use case?

Shouldn't there be a configuration with a subnet from where to get IPs?

plugins/IPv4/plugin.go
```diff
@@ -0,0 +1,297 @@
...
+		if netmask.To16() == nil {
+			return Handler6, Handler4, errors.New("plugins/IPv4: expected an IPv4 address, got: " + args[2])
+		}
+		if netmask.IsUnspecified() {
+			return Handler6, Handler4, errors.New("plugins/IPv4: netmask can not be 0.0.0.0, got: " + args[2])
+		}
+		if !checkValidNetmask(netmask) {
+			return Handler6, Handler4, errors.New("plugins/IPv4: netmask is not valid, got: " + args[2])
+		}
+		DNSServer = net.ParseIP(args[3])
```
pmazzini: You are putting all the logic in one file. The idea of CoreDHCP is to make it as modular as possible.

In this case the best approach is to include the DNS plugin which will add the DNS option to the replies and remove all the DNS logic from here.

README.md
```diff
@@ -23,7 +23,9 @@ server6:
         # - dns: 8.8.8.8 8.8.4.4 2001:4860:4860::8888 2001:4860:4860::8844
 
 #server4:
-#    listen: '127.0.0.1:67'
+#    listen: '0.0.0.0:67'
+#    plugins:
+         # - IPv4: leases4.txt 10.10.10.1 255.255.255.0 8.8.8.8 60 255.255.255.0
```
pmazzini: How about using `10.10.10.0/24` here instead of `10.10.10.1 255.255.255.0`.

Then you can use https://golang.org/pkg/net/#ParseCIDR

pmazzini:
I don't like that the name of the plugin is `IPv4` and has all the logic embedded into it. I think this can be done in a more modular way.

For example the IP assignment, reading and saving to the leases file can simply be done by modifying the `file` plugin, which is lacking all the features that you are adding for v4.

Then, adding the DNS server to the reply can be achieved by using the DNS plugin.

plugins/IPv4/plugin.go
pmazzini: Oh, I noticed there is no DNS plugin. One should be created.

plugins/DNS/plugin.go
```diff
@@ -0,0 +1,53 @@
+package dns
```
pmazzini: I would week the directory name lowercase to be consistent with the others. Can you change it from `plugins/DNS` to `plugins/dns`?

plugins/DNS/plugin.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/insomniacslk/dhcp/dhcpv6"
+)
+
+var (
+	dnsServers []net.IP
+)
+var log = logger.GetLogger()
+
+func init() {
+	plugins.RegisterPlugin("DNS", setupDNS6, setupDNS4)
```
pmazzini: I would keep the plugin name also in lowercase.

plugins/DNS/plugin.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/coredhcp/coredhcp/handler"
+	"github.com/coredhcp/coredhcp/logger"
+	"github.com/coredhcp/coredhcp/plugins"
+	"github.com/insomniacslk/dhcp/dhcpv4"
+	"github.com/insomniacslk/dhcp/dhcpv6"
+)
+
+var (
+	dnsServers []net.IP
```
pmazzini: Can this variable be moved below the init function to make it consistent with other plugins?

pmazzini:
I added a couple of NITs. Other than that the DNS plugin looks good.

README.md
pmazzini: I think this example is out of date.

cmds/coredhcp/main.go
pmazzini: My previous comment still stands here.

plugins/server/plugin.go
```diff
@@ -0,0 +1,111 @@
...
+	return resp, true
+}
+
+// Handler4 handles DHCPv4 packets for the file plugin
+func Handler4(req, resp *dhcpv4.DHCPv4) (*dhcpv4.DHCPv4, bool) {
+	if req.OpCode != dhcpv4.OpcodeBootRequest {
+		log.Printf("plugins/server: Not a BootRequest!")
+	}
+	switch mt := req.MessageType(); mt {
+	case dhcpv4.MessageTypeDiscover:
```
pmazzini: I wouldn't have a server plugin. I would completely remove this.

* The type of the reply message should be moved to the `MainHandler`: https://github.com/coredhcp/coredhcp/blob/master/coredhcp.go#L165

* The subnet mask can be set in the file plugin since there we define a subnet (or it can be a separate plugin).

* The server identifier can be set in the main handler if you want to automatically get the local IP address of where CoreDHCP is running or if you want to make it configurable you can use the `server_id` plugin

* The router can be a separate plugin if you want to make it configurable or it can be set in the file plugin if you want to use the 1st address of the subnet defined there.

pmazzini:
Thanks for you work so far @borna-blazevic!

plugins/file/plugin.go
```diff
@@ -113,10 +176,33 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
+		}
+		DHCPv4Records[req.ClientHWAddr.String()] = record
+
+	}
+	ipaddr := record.IP
+	log.Printf("plugins/file: found IP address %s for MAC %s", ipaddr, req.ClientHWAddr.String())
+	if req == nil {
+		log.Printf("plugins/file: Packet is nil!")
+	}
+	reply, err := dhcpv4.NewReplyFromRequest(req, dhcpv4.WithYourIP(ipaddr), dhcpv4.WithLeaseTime(LeaseTime))
```
pmazzini: You shouldn't create a `NewReplyFromRequest` here. It is already created in the `MainHandler`: https://github.com/coredhcp/coredhcp/blob/master/coredhcp.go#L175

plugins/file/plugin.go
```diff
@@ -125,24 +211,143 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+		DHCPv4Records = records
+		log.Printf("plugins/file: loaded %d leases from %s", len(DHCPv4Records), filename)
 	}
-	filename := args[0]
-	if filename == "" {
-		return nil, nil, errors.New("plugins/file: got empty file name")
+
+	return Handler6, Handler4, nil
+}
+func createIP(serverIP net.IP, netmask net.IP) (Record, bool) {
```
pmazzini: Is the boot returned in this function to indicate if an IP was allocated? It is better to return an error instead of a bool.

cmds/coredhcp/main.go
pmazzini: This comment still stands. Is `_ "github.com/coredhcp/coredhcp/plugins/example"` needed?

coredhcp.go
```diff
@@ -177,6 +177,9 @@ func (s *Server) MainHandler4(conn net.PacketConn, peer net.Addr, req *dhcpv4.DH
 		log.Printf("MainHandler4: failed to build reply: %v", err)
 		return
 	}
+	if req.OpCode != dhcpv4.OpcodeBootRequest {
```
pmazzini: This is not needed, it is already checked in line 171.

coredhcp.go
```diff
@@ -187,11 +190,21 @@ func (s *Server) MainHandler4(conn net.PacketConn, peer net.Addr, req *dhcpv4.DH
 	}
 
 	if resp != nil {
+		switch mt := req.MessageType(); mt {
```
pmazzini: I would move this a little bit up, to line 180, right after `NewReplyFromRequest` and before calling all the handlers.

plugins/file/plugin.go
```diff
@@ -21,16 +26,74 @@ func init() {
...
+			return nil, fmt.Errorf("plugins/file: expected an uint32, got: %v", ipaddr)
+		}
+		leased, err := strconv.ParseInt(tokens[3], 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("plugins/file: expected an uint32, got: %v", ipaddr)
+		}
+		if (leased + int64(leaseTime)) > time.Now().Unix() {
+			records[hwaddr.String()] = Record{IP: ipaddr, leaseTime: uint32(leaseTime), leased: leased}
+		}
+		err = saveRecords(records)
```
pmazzini: Why is it saving the records when reading them?

plugins/file/plugin.go
```diff
@@ -21,16 +26,74 @@ func init() {
...
+		tokens := strings.Fields(line)
+		if len(tokens) != 4 {
+			return nil, fmt.Errorf("plugins/file: malformed line: %s", line)
+		}
+		hwaddr, err := net.ParseMAC(tokens[0])
+		if err != nil {
+			return nil, fmt.Errorf("plugins/file: malformed hardware address: %s", tokens[0])
+		}
+		ipaddr := net.ParseIP(tokens[1])
+		if ipaddr.To16() == nil {
```
pmazzini: Why `To16`?

plugins/file/plugin.go
borna-blazevic: It is cleaning out the expired records.

cmds/coredhcp/main.go
borna-blazevic: > Is this line needed? I would leave the example plugin only as an example.

In this comment, you said you would leave it.

> This comment still stands. Is `_ "github.com/coredhcp/coredhcp/plugins/example"` needed?
It's not, but I left it as an example.

plugins/file/plugin.go
```diff
@@ -21,16 +26,74 @@ func init() {
...
 // DHCPv6Records and DHCPv4Records are mappings between MAC addresses in
 // form of a string, to network configurations.
 var (
 	DHCPv6Records map[string]net.IP
-	DHCPv4Records map[string]net.IP
+	DHCPv4Records map[string]Record
+	LeaseTime     uint32
+	filename      string
+	serverIP      net.IP
+	netmask       net.IP
```
pmazzini: Why not `net.IPMask` here?

cmds/coredhcp/main.go
pmazzini: I meat to say I would leave it as it was before :)

plugins/file/plugin.go
borna-blazevic: `net.IPMask ` does not have the functions necessary in `CreateIP` functions

plugins/file/plugin.go
```diff
@@ -125,24 +212,147 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+		if filename == "" {
+			return nil, nil, errors.New("plugins/file: got empty file name")
+		}
+		ip, network, err := net.ParseCIDR(args[1])
+		if err != nil {
+			return Handler6, Handler4, errors.New("plugins/file: expected an IPv4 address, got: " + args[1])
+		}
+		serverIP = ip
+
+		netmask = net.IPv4(network.Mask[0], network.Mask[1], network.Mask[2], network.Mask[3])
```
pmazzini: Why not `netmask = network.Mask`?

plugins/file/plugin.go
pmazzini: > `net.IPMask` does not have the functions necessary in `CreateIP` functions

Which functions are those?

plugins/file/plugin.go
```diff
@@ -125,24 +212,147 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+
+		if !checkValidNetmask(netmask) {
```
pmazzini: I think this is not needed, the `ParseCIDR` should return a valid netmask.

plugins/file/plugin.go
borna-blazevic: Nevermind i fixed it.

plugins/file/plugin.go
```diff
@@ -113,10 +176,34 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
-	//      and stop processing.
-	return resp, true
+	record, ok := DHCPv4Records[req.ClientHWAddr.String()]
+	if !ok {
+		log.Printf("plugins/file: MAC address %s is new, leasing new IP address", req.ClientHWAddr.String())
+		rec, err := createIP(serverIP, netmask)
+		if err != nil {
+			log.Error(err)
+			return nil, true
+
```
pmazzini: This extra line is not needed.

plugins/file/plugin.go
```diff
@@ -113,10 +176,34 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
+		}
+		err = saveIPAddress(rec, req.ClientHWAddr)
+		if err != nil {
+			log.Printf("plugins/file: SaveIPAddress failed: %v", err)
+		}
+		DHCPv4Records[req.ClientHWAddr.String()] = rec
+		record = rec
+	}
+	ipaddr := record.IP
```
pmazzini: This variable can be removed:

```
ipaddr := record.IP
resp.YourIPAddr = ipaddr
```

to 

```
resp.YourIPAddr = record.IP
```

plugins/file/plugin.go
```diff
@@ -113,10 +176,34 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
+	resp.YourIPAddr = ipaddr
+	dur, _ := time.ParseDuration(strconv.FormatUint(uint64(LeaseTime), 10) + "s")
+	resp.Options.Update(dhcpv4.OptIPAddressLeaseTime(dur))
+	subnet := netmask.To4()
+	resp.UpdateOption(dhcpv4.OptSubnetMask(net.IPv4Mask(subnet[0], subnet[1], subnet[2], subnet[3])))
+	resp.UpdateOption(dhcpv4.OptRouter(serverIP))
+	log.Printf("plugins/file: found IP address %s for MAC %s", ipaddr, req.ClientHWAddr.String())
+	if req == nil {
```
pmazzini: I don't think this check is needed.

plugins/file/plugin.go
pmazzini: Ok, I see, can a comment be added?

borna-blazevic: Sure

cmds/coredhcp/config.yml.example
```diff
@@ -2,12 +2,14 @@ server6:
     interface: "eth0"
     #listen: 'your-unicast-address:547'
     plugins:
-        - server_id: LL 00:de:ad:be:ef:00
+        #- server_id: LL 00:de:ad:be:ef:00
```
pmazzini: Are you changing this for any particular reason?

coredhcp.go
```diff
@@ -177,6 +177,14 @@ func (s *Server) MainHandler4(conn net.PacketConn, peer net.Addr, req *dhcpv4.DH
 		log.Printf("MainHandler4: failed to build reply: %v", err)
 		return
 	}
+	switch mt := req.MessageType(); mt {
+	case dhcpv4.MessageTypeDiscover:
+		tmp.UpdateOption(dhcpv4.OptMessageType(dhcpv4.MessageTypeOffer))
+	case dhcpv4.MessageTypeRequest:
+		tmp.UpdateOption(dhcpv4.OptMessageType(dhcpv4.MessageTypeAck))
+	default:
+		log.Printf("plugins/server: Unhandled message type: %v", mt)
```
pmazzini: I think it should `return` after logging this message.

cmds/coredhcp/config.yml.example
borna-blazevic: No, it was a mistake.

coredhcp.go
```diff
@@ -187,11 +195,14 @@ func (s *Server) MainHandler4(conn net.PacketConn, peer net.Addr, req *dhcpv4.DH
 	}
 
 	if resp != nil {
+
```
pmazzini: This extra line is not needed.

plugins/file/plugin.go
```diff
@@ -125,24 +209,137 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
-		return nil, nil, errors.New("plugins/file: got empty file name")
+
+	return Handler6, Handler4, nil
+}
+func createIP(serverIP net.IP, netmask net.IPMask) (Record, error) {
+
+	rand.Seed(time.Now().Unix())
+	ipserver := serverIP.To4()
+
+	ip := net.IPv4(random(1, 254), random(1, 254), random(1, 254), random(1, 254)).To4()
```
pmazzini: What is the logic behind this? Why random and not start from the first IP that you are able to allocate?

plugins/file/plugin.go
```diff
@@ -125,24 +209,137 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+		nextIP := make([]byte, 4)
+		binary.BigEndian.PutUint32(ip, ipInt)
+		for i := 0; i < 4; i++ {
+			nextIP[i] = (ip[i] & (netmask[i] ^ 255)) | (ipserver[i] & netmask[i])
+		}
+		if nextIP[0] != ip[0] || nextIP[1] != ip[1] || nextIP[2] != ip[2] || nextIP[3] != ip[3] {
+			break
+		}
+		ip = nextIP
+		taken = checkIfTaken(net.IPv4(ip[0], ip[1], ip[2], ip[3]))
```
pmazzini: Why `checkIfTaken(net.IPv4(ip[0], ip[1], ip[2], ip[3]))` instead of `checkIfTaken(ip)`

plugins/file/plugin.go
```diff
@@ -125,24 +209,137 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+		nextIP := make([]byte, 4)
+		binary.BigEndian.PutUint32(ip, ipInt)
+		for i := 0; i < 4; i++ {
+			nextIP[i] = (ip[i] & (netmask[i] ^ 255)) | (ipserver[i] & netmask[i])
+		}
+		if nextIP[0] != ip[0] || nextIP[1] != ip[1] || nextIP[2] != ip[2] || nextIP[3] != ip[3] {
+			return Record{}, errors.New("plugins/file: no new IP addresses available")
+		}
+		ip = nextIP
+		taken = checkIfTaken(net.IPv4(ip[0], ip[1], ip[2], ip[3]))
```
pmazzini: Why `checkIfTaken(net.IPv4(ip[0], ip[1], ip[2], ip[3]))` instead of `checkIfTaken(ip)`?

plugins/file/plugin.go
borna-blazevic: No particular reason, is this an issue?

plugins/file/plugin.go
```diff
@@ -125,24 +209,137 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+
+	}
+	return Record{IP: net.IPv4(ip[0], ip[1], ip[2], ip[3]), leaseTime: LeaseTime, leased: time.Now().Unix()}, nil
```
pmazzini: Why `Record{IP: net.IPv4(ip[0], ip[1], ip[2], ip[3]), leaseTime: LeaseTime, leased: time.Now().Unix()}, nil` instead of `Record{IP: ip), leaseTime: LeaseTime, leased: time.Now().Unix()}, nil`?

plugins/file/plugin.go
borna-blazevic: This way the code can be expanded, so the initial IP is generated based on the MAC of the device later on.

plugins/file/plugin.go
```diff
@@ -125,24 +209,137 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
 	}
-	filename := args[0]
-	if filename == "" {
-		return nil, nil, errors.New("plugins/file: got empty file name")
+
+	return Handler6, Handler4, nil
+}
+func createIP(serverIP net.IP, netmask net.IPMask) (Record, error) {
+
+	rand.Seed(time.Now().Unix())
```
pmazzini: AFAIK the seed should be generated once and not on every call to `createIP`

plugins/file/plugin.go
```diff
@@ -125,24 +209,137 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+	ipserver := serverIP.To4()
```
pmazzini: Is the `To4()` needed?

plugins/file/plugin.go
pmazzini: Is the `To4()` needed?

plugins/file/plugin.go
```diff
@@ -21,16 +26,76 @@ func init() {
 	plugins.RegisterPlugin("file", setupFile6, setupFile4)
 }
 
+//Record holds an IP lease record
+type Record struct {
+	IP        net.IP
+	leaseTime uint32
```
pmazzini: This represents a duration, how about using a int64 or time.Duration. https://golang.org/pkg/time/#Duration

plugins/file/plugin.go
```diff
@@ -125,24 +209,137 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+		ip[i] = (ip[i] & (netmask[i] ^ 255)) | (ipserver[i] & netmask[i])
 	}
-	records, err := LoadDHCPv6Records(filename)
+	taken := checkIfTaken(net.IPv4(ip[0], ip[1], ip[2], ip[3]))
+	for taken {
+		ipInt := binary.BigEndian.Uint32(ip)
+		ipInt++
+		nextIP := make([]byte, 4)
+		binary.BigEndian.PutUint32(ip, ipInt)
+		for i := 0; i < 4; i++ {
```
pmazzini: If I understand correctly, here you are checking if the next IP is in the subnet. Why not use https://golang.org/pkg/net/#IPNet.Contains?

plugins/file/plugin.go
```diff
@@ -21,16 +26,78 @@ func init() {
...
+		leaseTime, err := time.ParseDuration(tokens[2])
+		if err != nil {
+			return nil, fmt.Errorf("plugins/file: expected an uint32, got: %v", ipaddr)
+		}
+		leased, err := strconv.ParseInt(tokens[3], 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("plugins/file: expected an uint32, got: %v", ipaddr)
+		}
+		//Only if the record is not expired.
+		println()
```
pmazzini: Are these two `println` going to be removed?

plugins/file/plugin.go
```diff
@@ -113,10 +180,28 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
+		}
+		err = saveIPAddress(rec, req.ClientHWAddr)
+		if err != nil {
+			log.Printf("plugins/file: SaveIPAddress failed: %v", err)
+		}
+		DHCPv4Records[req.ClientHWAddr.String()] = rec
+		record = rec
+	}
+	resp.YourIPAddr = record.IP
+	dur, _ := time.ParseDuration(strconv.FormatUint(uint64(LeaseTime), 10) + "s")
```
pmazzini: LeaseTime is now a `time.Duration`. This line can be simplified.

plugins/file/plugin.go
borna-blazevic: Yes, my bad.

plugins/file/plugin.go
pmazzini: Or even removed.

borna-blazevic: Yes, missed it.

plugins/file/plugin.go
```diff
@@ -125,24 +210,121 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+	}
+	for taken {
+		ipInt := binary.BigEndian.Uint32(ip)
+		ipInt--
+		binary.BigEndian.PutUint32(ip, ipInt)
+		if !server.Contains(ip) {
+			return Record{}, errors.New("plugins/file: no new IP addresses available")
+		}
+		taken = checkIfTaken(ip)
+
```
pmazzini: I would remove this extra line.

plugins/file/plugin.go
```diff
@@ -21,16 +26,78 @@ func init() {
...
 
 // DHCPv6Records and DHCPv4Records are mappings between MAC addresses in
 // form of a string, to network configurations.
 var (
 	DHCPv6Records map[string]net.IP
-	DHCPv4Records map[string]net.IP
+	DHCPv4Records map[string]Record
+	LeaseTime     time.Duration
+	filename      string
+	server        net.IPNet
```
pmazzini: The variable name `server` is not descriptive in this case because you are storing a subnet. Can it be changed to `network` or something similar? Also, can it be a reference instead of the whole struct?

plugins/file/plugin.go
pmazzini: This comment still stands.

plugins/file/plugin.go
borna-blazevic: Sure

plugins/file/plugin.go
```diff
@@ -113,10 +177,27 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
+		err = saveIPAddress(rec, req.ClientHWAddr)
+		if err != nil {
+			log.Printf("plugins/file: SaveIPAddress failed: %v", err)
+		}
+		DHCPv4Records[req.ClientHWAddr.String()] = rec
+		record = rec
+	}
+	resp.YourIPAddr = record.IP
+	resp.Options.Update(dhcpv4.OptIPAddressLeaseTime(LeaseTime))
+	resp.UpdateOption(dhcpv4.OptSubnetMask((*network).Mask))
```
pmazzini: There is no need to do a dereference for the network here.

plugins/file/plugin.go
```diff
@@ -113,10 +177,27 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
+	resp.UpdateOption(dhcpv4.OptRouter((*network).IP))
```
pmazzini: Same as above.

plugins/file/plugin.go
```diff
@@ -125,24 +206,120 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+
+		log.Printf("plugins/file: loaded %d leases from %s", len(DHCPv4Records), filename)
+	}
+
+	return Handler6, Handler4, nil
+}
+func createIP(network *net.IPNet) (Record, error) {
+	ip := []byte{random(1, 254), random(1, 254), random(1, 254), random(1, 254)}
+	for i := 0; i < 4; i++ {
+		ip[i] = (ip[i] & ((*network).Mask[i] ^ 255)) | ((*network).IP[i] & (*network).Mask[i])
```
pmazzini: There is no need to do a dereference of the network.

plugins/file/plugin.go
```diff
@@ -125,24 +206,120 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+		if !(*network).Contains(ip) {
+			break
+		}
+		taken = checkIfTaken(ip)
+	}
+	for taken {
+		ipInt := binary.BigEndian.Uint32(ip)
+		ipInt--
+		binary.BigEndian.PutUint32(ip, ipInt)
+		if !(*network).Contains(ip) {
```
pmazzini: Same here.

plugins/file/plugin.go
```diff
@@ -125,24 +206,120 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+	ip := []byte{random(1, 254), random(1, 254), random(1, 254), random(1, 254)}
+	for i := 0; i < 4; i++ {
+		ip[i] = (ip[i] & ((*network).Mask[i] ^ 255)) | ((*network).IP[i] & (*network).Mask[i])
+	}
+	taken := checkIfTaken(ip)
+	for taken {
+		ipInt := binary.BigEndian.Uint32(ip)
+		ipInt++
+		binary.BigEndian.PutUint32(ip, ipInt)
+		if !(*network).Contains(ip) {
```
pmazzini: Same here.

plugins/dns/plugin.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func setupDNS4(args ...string) (handler.Handler4, error) {
+	log.Printf("plugins/dns: loaded plugin for DHCPv4.")
+	if len(args) < 1 {
+		return nil, errors.New("plugins/dns: need at least one DNS server")
+	}
+	for _, arg := range args {
+		DNSServer := net.ParseIP(arg)
+		if DNSServer.To16() == nil {
```
pmazzini: Why the `To16`?

plugins/file/plugin.go
```diff
@@ -125,24 +206,120 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+		DHCPv4Records = records
+
+		rand.Seed(time.Now().Unix())
+
+		log.Printf("plugins/file: loaded %d leases from %s", len(DHCPv4Records), filename)
+	}
+
+	return Handler6, Handler4, nil
+}
+func createIP(network *net.IPNet) (Record, error) {
```
pmazzini: I think ti is better if this returns a reference to a Record and avoid copying the whole struct.

plugins/file/plugin.go
```diff
@@ -125,24 +206,120 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
-	records, err := LoadDHCPv6Records(filename)
+	for _, v := range DHCPv4Records {
+		if v.IP.String() == ip.String() && (v.leased+int64(v.leaseTime.Seconds()) > time.Now().Unix()) {
+			taken = true
+			break
+		}
+	}
+	return taken
+}
+func saveIPAddress(record Record, mac net.HardwareAddr) error {
```
pmazzini: I think ti is better if this takes a reference to a Record and avoid copying the whole struct.

insomniacslk:
On mobile and on holidays, so I can't review properly, but why are file permissions being updated from 0644 to 0755?

plugins/file/plugin.go
```diff
@@ -113,10 +177,27 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
+		if err != nil {
+			log.Printf("plugins/file: SaveIPAddress failed: %v", err)
+		}
+		DHCPv4Records[req.ClientHWAddr.String()] = rec
+		record = rec
+	}
+	resp.YourIPAddr = record.IP
+	resp.Options.Update(dhcpv4.OptIPAddressLeaseTime(LeaseTime))
+	resp.UpdateOption(dhcpv4.OptSubnetMask(network.Mask))
+	resp.UpdateOption(dhcpv4.OptRouter(network.IP))
```
pmazzini: `network.IP` is something like `10.10.10.0` so this is probably not what you want to set as router.

pmazzini:
The changes to coredhcp, server_id, dns and router plugin look good to me.

I think the file plugin still needs some changes.

plugins/router/plugin.go
```diff
@@ -0,0 +1,54 @@
+package router
```
pmazzini: Just out of curiosity, are you using this plugin? I don't see it listed in the config.

plugins/file/plugin.go
```diff
@@ -21,16 +26,75 @@ func init() {
 	plugins.RegisterPlugin("file", setupFile6, setupFile4)
 }
 
+//Record holds an IP lease record
+type Record struct {
+	IP        net.IP
```
pmazzini: Can't this be simplified to contain only the timestamp of when the lease expires instead of when the lease was assigned plus its duration?

plugins/file/plugin.go
```diff
@@ -125,24 +205,120 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+func saveIPAddress(record *Record, mac net.HardwareAddr) error {
+	f, err := os.OpenFile(filename, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
 	if err != nil {
-		return nil, nil, fmt.Errorf("plugins/file: failed to load DHCPv6 records: %v", err)
+		return err
 	}
-	log.Printf("plugins/file: loaded %d leases from %s", len(records), filename)
-	StaticRecords = records
+	defer f.Close()
+	_, err = f.WriteString(mac.String() + " " + record.IP.String() + " " + strconv.FormatUint(uint64(record.leaseTime.Seconds()), 10) + "s " + strconv.FormatInt(record.leased, 10) + "\n")
```
pmazzini: Same as above, can't this be simplified to contain only the timestamp of when the lease expires instead of when the lease was assigned plus its duration?

plugins/file/plugin.go
```diff
@@ -125,24 +205,120 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
-	records, err := LoadDHCPv6Records(filename)
+	for _, v := range DHCPv4Records {
+		if v.IP.String() == ip.String() && (v.leased+int64(v.leaseTime.Seconds()) > time.Now().Unix()) {
+			taken = true
+			break
+		}
+	}
+	return taken
+}
+func saveIPAddress(record *Record, mac net.HardwareAddr) error {
```
pmazzini: NIT: can the first parameter of the function be the mac since that is what we are using as index.

cmds/coredhcp/config.yml.example
```diff
@@ -9,5 +9,7 @@ server6:
 server4:
     listen: '0.0.0.0:67'
     interface: "eth0"
-    plugins:
-        - server_id: 192.168.1.12
+    plugins:        
+        #- server_id: 10.10.10.1
+        #- file: leases.txt 10.10.10.1/24 60
```
pmazzini: After giving it a though, I am not sure this is the proper format we should use to define a pool of IPs.
For instance, if you define `10.10.10.1/24` the DHCP server may assign a client the IP `10.10.10.1` which you may want to use as a static IP for a router.

I think that instead this should have a starting IP and an ending IP for the pool. Ex. `10.10.10.100 10.10.10.200`.

In that way you can avoid having to use the whole subnet for dynamic IPs. Thoughts?

pmazzini: And we can have the `netmask` as another plugin/option.

plugins/router/plugin.go
borna-blazevic: Yes, i forgot to put it in the example config

cmds/coredhcp/config.yml.example
borna-blazevic: There was a check so the address of the router is never leased, but now I think it is not doing what it's supposed to do because the serverIP is not the same as the router IP. Yes it can be implemented so a range of IPs is set. Do you think file plugin arguments should look like:
`file: leases.txt 10.10.10.1/24 10.10.10.100 10.10.10.200  60`
or
`file: leases.txt 10.10.10.1 10.10.10.100 10.10.10.200  60`

plugins/file/plugin.go
borna-blazevic: Sure

cmds/coredhcp/config.yml.example
pmazzini: I think it should be either like this

```
file: leases.txt 10.10.10.100 10.10.10.200 60
```

There is no need to add `10.10.10.1/24` or `10.10.10.1`.

Or even like this:

```
file: leases.txt 10.10.10.100 10.10.10.200
```

I wonder if we can decouple the lease time as well.

plugins/router/plugin.go
pmazzini: Again, just out of curiosity: Does your router have a different IP to the one of the DHCP server?

plugins/file/plugin.go
```diff
@@ -21,16 +26,70 @@ func init() {
...
+
 // StaticRecords holds a MAC -> IP address mapping
 var StaticRecords map[string]net.IP
 
 // DHCPv6Records and DHCPv4Records are mappings between MAC addresses in
 // form of a string, to network configurations.
 var (
 	DHCPv6Records map[string]net.IP
-	DHCPv4Records map[string]net.IP
+	DHCPv4Records map[string]Record
```
pmazzini: Can this be a map of a reference to `Record` instead?

plugins/router/plugin.go
borna-blazevic: No, it doesn't, but I think it can have a different IP to the one of the DHCP server.

cmds/coredhcp/config.yml.example
borna-blazevic: > I wonder if we can decouple the lease time as well.

How would the file plugin save the leases with the time of expiry?
> ```
> file: leases.txt 10.10.10.100 10.10.10.200
> ```
How would I get the serverIP from here?

plugins/file/plugin.go
borna-blazevic: Yes.

cmds/coredhcp/config.yml.example
pmazzini: > How would I get the serverIP from here?

What do you need a `serverIP` for in the `file` plugin?

plugins/router/plugin.go
pmazzini: Do you need to set the `server_id` and the `router` in that case?

borna-blazevic: My understanding of it is that the router field lists the available routers the client can connect to, while the server_id is used so the server knows it is supposed to respond to the DHCP request. So yes the server_id is always necessary.

cmds/coredhcp/config.yml.example
borna-blazevic: I don't need the server IP only if I start at the bottom range and continue to increment the IP until a free lease is found. The way it's implemented now is a random IP (which can be later a hash function so the same device always gets the same initial IP ), and then it is adjusted so it fits the netmask. To do that it uses the server IP and the server netmask.
`for i := 0; i < 4; i++ {`
		`ip[i] = (ip[i] & (network.Mask[i] ^ 255)) | (network.IP[i] & network.Mask[i])`
`}`

plugins/router/plugin.go
pmazzini: I thought the `server_id` plugin was also setting the `GatewayIPAddr`: https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/dhcpv4.go#L69.

Disregards.

cmds/coredhcp/config.yml.example
pmazzini: If you want to randomize you can still do it, you can pick an IP from the pool. You only need to use a different logic.

plugins/file/plugin.go
```diff
@@ -21,16 +26,63 @@ func init() {
...
 
 // DHCPv6Records and DHCPv4Records are mappings between MAC addresses in
 // form of a string, to network configurations.
 var (
 	DHCPv6Records map[string]net.IP
-	DHCPv4Records map[string]net.IP
+	DHCPv4Records map[string]*Record
+	LeaseTime     time.Duration
+	filename      string
+	//network       *net.IPNet
```
pmazzini: I would remove this comment

plugins/netmask/plugin.go
```diff
@@ -0,0 +1,67 @@
...
+	netmaskIP := net.ParseIP(args[0])
+	if netmaskIP.IsUnspecified() {
+		return nil, errors.New("plugins/file: netmask is not valid, got: " + args[1])
+	}
+	netmaskIP = netmaskIP.To4()
+	if netmaskIP == nil {
+		return nil, errors.New("plugins/file: expected an netmask address, got: " + args[1])
+	}
+	netmask = net.IPv4Mask(netmaskIP[0], netmaskIP[1], netmaskIP[2], netmaskIP[3])
+	println(netmask.String())
```
pmazzini: Is this `println` a left over?

pmazzini:
Other than that it looks good but I'll have another look tomorrow.

plugins/file/plugin.go
```diff
@@ -113,10 +165,25 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
 // Handler4 handles DHCPv4 packets for the file plugin
 func Handler4(req, resp *dhcpv4.DHCPv4) (*dhcpv4.DHCPv4, bool) {
-	// TODO check the MAC address in the request
-	//      if it is present in StaticRecords, forge a response
-	//      and stop processing.
-	return resp, true
+	record, ok := DHCPv4Records[req.ClientHWAddr.String()]
+	if !ok {
+		log.Printf("plugins/file: MAC address %s is new, leasing new IP address", req.ClientHWAddr.String())
+		rec, err := createIP(ipRangeStart, ipRangeEnd)
```
insomniacslk: this means that the IP address is picked in a range, while the `file` plugin is meant for static file mappings. This should be implemented in a different plugin, e.g. something called `range` or similar name.

plugins/file/plugin.go
```diff
@@ -125,24 +192,124 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
+		taken = checkIfTaken(ip)
 	}
-	filename := args[0]
-	if filename == "" {
-		return nil, nil, errors.New("plugins/file: got empty file name")
+	return &Record{IP: ip, expires: time.Now().Unix() + int64(LeaseTime.Seconds())}, nil
+
+}
+func random(min uint32, max uint32) uint32 {
+	return rand.Uint32()%(max-min+1) + min
```
insomniacslk: this function suffers of modulo bias, you should use `rand.Intn` instead, https://golang.org/pkg/math/rand/#Intn

plugins/file/plugin.go
```diff
@@ -21,16 +26,63 @@ func init() {
 	plugins.RegisterPlugin("file", setupFile6, setupFile4)
 }
 
+//Record holds an IP lease record
+type Record struct {
+	IP      net.IP
+	expires int64
```
insomniacslk: shouldn't this be a `time.Time` ?

plugins/netmask/plugin.go
```diff
@@ -0,0 +1,67 @@
...
+func init() {
+	plugins.RegisterPlugin("netmask", setupNetmask6, setupNetmask4)
+}
+
+var (
+	netmask net.IPMask
+)
+
+func setupNetmask6(args ...string) (handler.Handler6, error) {
+	log.Printf("plugins/netmask: loaded plugin for DHCPv6.")
```
insomniacslk: please add a TODO/Warningf/issue here to say that this is not implemented yet for v6

plugins/file/plugin.go
borna-blazevic: If I were to implement another plugin called range that creates a new IP and leases it, that plugin would still have to open the leases file, check if the IP is taken and save it after it was leased. Then the file plugin and the range plugin would be very similar. Is that something you want? If so, then the range plugin would accept the arguments the file plugin accepts now, and the file plugin would accept the filename and a lease-time so it can expand the lease if it expired once it leases an IP address? Furthermore, file plugin would be second to last in the order of plugins and return false if it found an existing IP, and the range plugin would be the last plugin and would handle the request only if the file plugin returned false?

plugins/file/plugin.go
```diff
@@ -125,24 +191,124 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
-		return nil, nil, fmt.Errorf("plugins/file: failed to load DHCPv6 records: %v", err)
+		return err
 	}
-	log.Printf("plugins/file: loaded %d leases from %s", len(records), filename)
-	StaticRecords = records
+	return nil
+}
 
-	return Handler6, Handler4, nil
+func saveRecords(DHCPv4Records map[string]*Record) error {
```
pmazzini: After the latest changes this function is not being called.

plugins/file/plugin.go
pmazzini: I don't have a strong opinion about it. The plugins will share some logic. Maybe we can have them separate for now, and at some point if it makes sense we can merge them

insomniacslk: As long as the plug-in behaviour is consistent (i.e. v4 and v6 do the same things where it makes sense) and the name and documentation match what the plug-in does, I'm fine with it.
In this case the file plugin is for static file mapping (like a database), and meant to be read only. I'm ok with changing it (and renaming) to something more useful in the real world, like the range plugin discussed above, and if we do so, the v6 code has to behave the same way.

To sum up, what about renaming it to `ramge` and do the same with v6?

Regarding static mapping, I had in mind writing a sqlite plugin for quite some time, that could do what `file` does today before your change

README.md
```diff
@@ -23,7 +23,14 @@ server6:
...
 
 #server4:
-#    listen: '127.0.0.1:67'
+#    listen: '0.0.0.0:67'
+#    plugins:
+        # - server_id: 10.10.10.1
+        # - dns: 8.8.8.8 8.8.4.4
+        # - router: 10.10.10.1
+        # - netmask: 255.255.255.0
+        # - file: leases.txt 60s
```
pmazzini: I think it is better to remove this from the example. I would use either one or the other plugin.

cmds/coredhcp/config.yml.example
```diff
@@ -10,4 +10,9 @@ server4:
     listen: '0.0.0.0:67'
     interface: "eth0"
     plugins:
-        - server_id: 192.168.1.12
+        # - server_id: 10.10.10.1
+        # - dns: 8.8.8.8 8.8.4.4
+        # - router: 10.10.10.1
+        # - netmask: 255.255.255.0
+        # - file: leases.txt 60s
```
pmazzini: Same as above, I would use either `file` or `range` plugin. I would remove this line from the config.

plugins/file/plugin.go
```diff
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net"
 	"strings"
+	"time"
```
pmazzini: I would leave the `file` plugin unmodified, and simply focus on the `range` one.

borna-blazevic: Do you mean reverting file plugin to its original state?

borna-blazevic: With an empty handler for ipv4?

pmazzini: Yes, I would leave it as it was before, without the lease time config in it since it is aimed for static allocation. There is not much sense to store the lease time in the file.

pmazzini:
Aside from the last comments it looks good to me. Thanks @borna-blazevic!

plugins/file/plugin.go
pmazzini: You can add feature parity with the v6 one but I wouldn't add lease time to it.

borna-blazevic: So then the range plugin would have a different lease file and range and plugin should not be running on the same server?

borna-blazevic: If so, then the range plugin should check the leased IP addresses as file plugin did before this commit?

pmazzini: You can run them in the same server, I would use a different file for static and dynamic allocation. I would keep the plugins independent so that the file format change won't break both plugins.

borna-blazevic: > You can add feature parity with the v6 one but I wouldn't add lease time to it.

I'm not sure I understood you, could you elaborate?

borna-blazevic: > You can run them in the same server, I would use a different file for static and dynamic allocation. I would keep the plugins independent so that a file format change won't break both plugins.

But then it could happen that the same IP address is leased to different devices.

pmazzini: > I'm not sure I understood you, could you elaborate?

When I said "You can add feature parity with the v6 one but I wouldn't add lease time to it." I was replying to your "With an empty handler for ipv4?" question.

You can go ahead and modify the `file` plugin if you want to add the missing functionality but I wouldn't add a lease time to it. I would focus on the `range` plugin at this point.

borna-blazevic: Ok, I got it now, thank you.

pmazzini: > EDIT: I guess you could prevent that so the static IP addresses are not in the range of the range plugin.

Right, you can have a pool of IPs for each thing.

README.md
```diff
@@ -23,7 +23,13 @@ server6:
...
 
 #server4:
-#    listen: '127.0.0.1:67'
+#    listen: '0.0.0.0:67'
+#    plugins:
+        # - server_id: 10.10.10.1
+        # - dns: 8.8.8.8 8.8.4.4
+        # - router: 10.10.10.1
+        # - netmask: 255.255.255.0
+        # - file: leases.txt
```
pmazzini: Lets add the range plugin you will be using in the example. That is the one you will be using right? The other example already has the file plugin.

cmds/coredhcp/config.yml.example
```diff
@@ -10,4 +10,8 @@ server4:
     listen: '0.0.0.0:67'
     interface: "eth0"
     plugins:
-        - server_id: 192.168.1.12
+        # - server_id: 10.10.10.1
+        # - dns: 8.8.8.8 8.8.4.4
+        # - router: 10.10.10.1
+        # - netmask: 255.255.255.0
+        # - file: leases.txt
```
pmazzini: Lets add the range plugin you will be using in the example. That is the one you will be using right?

plugins/file/plugin.go
```diff
@@ -125,24 +162,41 @@ func setupFile6(args ...string) (handler.Handler6, error) {
...
-		return nil, nil, errors.New("plugins/file: need a file name")
-	}
-	filename := args[0]
-	if filename == "" {
-		return nil, nil, errors.New("plugins/file: got empty file name")
-	}
-	records, err := LoadDHCPv6Records(filename)
-	if err != nil {
-		return nil, nil, fmt.Errorf("plugins/file: failed to load DHCPv6 records: %v", err)
+	if v6 {
```
pmazzini: This can be simplified since v6 and v4 share some logic:

```
if len(args) < 1 {
        return nil, nil, errors.New("plugins/file: need a file name")
}
filename := args[0]
if filename == "" {
        return nil, nil, errors.New("plugins/file: got empty file name")
}
if v6 {
        records, err := LoadDHCPv6Records(filename)
} else {
        records, err := LoadDHCPv4Records(filename)
}
```

plugins/range/plugin.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// StaticRecords holds a MAC -> IP address mapping
+var StaticRecords map[string]net.IP
+
+// DHCPv6Records and DHCPv4Records are mappings between MAC addresses in
+// form of a string, to network configurations.
+var (
+	// TODO change DHCPv6Records to Record
+	DHCPv6Records map[string]net.IP
```
pmazzini: For consistency, even if it is not used, I would change

```
DHCPv6Records map[string]net.IP
```
with:
```
DHCPv6Records map[string]*Record
```

plugins/range/plugin.go
```diff
@@ -0,0 +1,220 @@
...
+			return nil, nil, errors.New("expected an IP address, got: " + args[2])
+		}
+		if binary.BigEndian.Uint32(ipRangeStart.To4()) >= binary.BigEndian.Uint32(ipRangeEnd.To4()) {
+			return nil, nil, errors.New("start of IP range has to be lower than the end fo an IP range")
+		}
+		var err error
+		LeaseTime, err = time.ParseDuration(args[3])
+		if err != nil {
+			return Handler6, Handler4, errors.New("expected an uint32, got: " + args[3])
+		}
```
pmazzini: I would do the `if v6 {` here since this is where it starts the v4 specific part.

plugins/range/plugin.go
```diff
@@ -0,0 +1,230 @@
...
+	return Handler6, nil
+}
+
+func setupRange4(args ...string) (handler.Handler4, error) {
+	_, h4, err := setupRange(false, args...)
+	return h4, err
+}
+
+func setupRange(v6 bool, args ...string) (handler.Handler6, handler.Handler4, error) {
+	if v6 {
```
pmazzini: I think this if can be removed, it is already in line 160.

pmazzini:
I think this is good to merge after addressing my last comment.

PR #42, opened 2019-08-25
Added plugin imports in cmds/coredhcp

insomniacslk:
New plugins have been added but not imported in cmds/coredhcp.


insomniacslk:
I think we can merge this despite the failure on `tip`, which can be fixed separately

PR #43, opened 2019-08-25
Logger: add prefix to log lines

insomniacslk:
This is useful to automatically provide context about what component of
the framework is emitting a log line.


insomniacslk:
CI failing on `tip` again, we can proceed despite that

PR #44, opened 2019-08-26
[plugins/range] Various fixes

insomniacslk:
* removed unused variables
* removed some code duplication
* various errors/logging fixes

### DIFF
--- a/plugins/range/plugin.go
+++ b/plugins/range/plugin.go
@@ -1,11 +1,11 @@
 package rangeplugin
 
 import (
-	"bytes"
+	"bufio"
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"math/rand"
 	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


insomniacslk:
CC @borna-blazevic

insomniacslk:
usual thing for `tip`

plugins/range/plugin.go
```diff
@@ -31,34 +32,26 @@ type Record struct {
...
-}
-
-// LoadDHCPv4Records loads the DHCPv4Records global map with records stored on
+// loadRecords loads the DHCPv6/v4 Records global map with records stored on
 // the specified file. The records have to be one per line, a mac address and an
-// IPv4 address.
-func LoadDHCPv4Records(filename string) (map[string]*Record, error) {
-	log.Printf("reading leases from %s", filename)
+// IP address.
+func loadRecords(r io.Reader, v6 bool) (map[string]*Record, error) {
```
pmazzini: The 1st parameter (io.Reader) is not being used.

plugins/range/plugin.go
```diff
@@ -133,41 +132,52 @@ func setupRange4(args ...string) (handler.Handler4, error) {
...
 	if binary.BigEndian.Uint32(ipRangeStart.To4()) >= binary.BigEndian.Uint32(ipRangeEnd.To4()) {
-		return nil, nil, errors.New("start of IP range has to be lower than the end fo an IP range")
+		return nil, nil, errors.New("start of IP range has to be lower than the end of an IP range")
 	}
 	LeaseTime, err = time.ParseDuration(args[3])
 	if err != nil {
-		return Handler6, Handler4, errors.New("expected an uint32, got: " + args[3])
+		return Handler6, Handler4, fmt.Errorf("invalid duration: %v", args[3])
+	}
+	r, err := os.Open(filename)
```
pmazzini: This file is not being closed and I think it is not being used either.

plugins/range/plugin.go
insomniacslk: fixed

plugins/range/plugin.go
insomniacslk: related to the above

plugins/range/plugin.go
insomniacslk: another indicator that using global variables is rarely a good idea...

plugins/range/plugin.go
```diff
@@ -31,41 +31,25 @@ type Record struct {
...
-	if err != nil {
-		return nil, err
-	}
+// IP address.
+func loadRecords(r io.Reader, v6 bool) (map[string]*Record, error) {
+	sc := bufio.NewScanner(r)
 	records := make(map[string]*Record)
-	for _, lineBytes := range bytes.Split(data, []byte{'\n'}) {
-		line := string(lineBytes)
+	for sc.Scan() {
```
pmazzini: Should the same thing be done in the `file` plugin for consistency? https://github.com/coredhcp/coredhcp/blob/master/plugins/file/plugin.go#L78

insomniacslk: yes, but in another PR, I don't want to stash too many changes here.

My plan is to move the commob stuff to `pkg`, and rebase `file` and  `range` on it

PR #53, opened 2019-09-16
server4: Fix peer address logic

Natolumin:
I noticed responding to relayed messages was broken on dhcpv4 (the response would go to broadcast, because of an `if` that should have been an `else if`

The RFC has this to say about how to select the address to respond to:

>  If the 'giaddr' field in a DHCP message from a client is non-zero,
>  the server sends any return messages to the 'DHCP server' port on the
>  BOOTP relay agent whose address appears in 'giaddr'. If the 'giaddr'
>  field is zero and the 'ciaddr' field is nonzero, then the server
>  unicasts DHCPOFFER and DHCPACK messages to the address in 'ciaddr'.
>  If 'giaddr' is zero and 'ciaddr' is zero, and the broadcast bit is
>  set, then the server broadcasts DHCPOFFER and DHCPACK messages to
>  0xffffffff. If the broadcast bit is not set and 'giaddr' is zero and
>  'ciaddr' is zero, then the server unicasts DHCPOFFER and DHCPACK
>  messages to the client's hardware address and 'yiaddr' address.  In
>  all cases, when 'giaddr' is zero, the server broadcasts any DHCPNAK
>  messages to 0xffffffff.

We can implement almost all of this. Unfortunately the bit about non-broadcast
responses ("If the broadcast bit is not set and 'giaddr' is zero and
'ciaddr' is zero, then the server unicasts DHCPOFFER and DHCPACK
messages to the client's hardware address and 'yiaddr' address")
is impossible to implement with the current API offered by go and the
dhcp library (it requires raw sockets and a way to pass L2 information
in the peer address).

The RFC also says this which saves us in the non-broadcast case:

> If unicasting is not possible, the message MAY be sent as an IP
> broadcast using an IP broadcast address (preferably 0xffffffff) as the
> IP destination address and the link-layer broadcast address as the
> link-layer destination address.

### DIFF
--- a/coredhcp.go
+++ b/coredhcp.go
@@ -162,7 +162,7 @@ func (s *Server) MainHandler6(conn net.PacketConn, peer net.Addr, req dhcpv6.DHC
 }
 
 // MainHandler4 is like MainHandler6, but for DHCPv4 packets.
-func (s *Server) MainHandler4(conn net.PacketConn, peer net.Addr, req *dhcpv4.DHCPv4) {
+func (s *Server) MainHandler4(conn net.PacketConn, _peer net.Addr, req *dhcpv4.DHCPv4) {
 	var (
 		resp, tmp *dhcpv4.DHCPv4
 		err       error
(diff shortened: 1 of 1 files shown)


Natolumin:
V2: silence the lint warning about `peer` being overwritten before use

PR #54, opened 2019-09-16
plugins/dns: Implement DHCPv6 support

Natolumin:
Patch 1 (trivially) implements sending DNS servers in DHCPv6 response messages
Patch 2 removes the hardcoded list in the file plugin
Patch 3 improves the logic of the plugin to respect Option requests in both protocols

There are unittests and I manually tested both in relayed and non-relayed modes of operation with dhclient


plugins/dns/plugin.go
```diff
@@ -37,20 +47,46 @@ func setupDNS4(args ...string) (handler.Handler4, error) {
...
+	oro := decap.GetOneOption(dhcpv6.OptionORO)
+	if oro == nil {
+		// A client MUST include an Option Request option in a Solicit, Request,
+		// Renew, Rebind, or Information-request message to inform the server
+		// about options the client wants the server to send to the client
+		log.Warn("No OptionRequestOption in request, dropping")
+		return nil, true
+	}
+
+	for _, opt := range oro.(*dhcpv6.OptRequestedOption).RequestedOptions() {
```
pmazzini: NIT: this can be replaced with the `IsOptionRequested` method:
https://github.com/insomniacslk/dhcp/blob/master/dhcpv6/dhcpv6message.go#L239

plugins/dns/plugin.go
```diff
@@ -37,20 +47,46 @@ func setupDNS4(args ...string) (handler.Handler4, error) {
...
+			break
+		}
+	}
 	return resp, false
 }
 
 //Handler4 handles DHCPv4 packets for the dns plugin
 func Handler4(req, resp *dhcpv4.DHCPv4) (*dhcpv4.DHCPv4, bool) {
-	resp.Options.Update(dhcpv4.OptDNS(dnsServers...))
+	if opts := req.ParameterRequestList(); opts != nil && !opts.Has(dhcpv4.OptionDomainNameServer) {
```
pmazzini: NIT: this can be replaced with the `IsOptionRequested` method:
https://github.com/insomniacslk/dhcp/blob/master/dhcpv4/dhcpv4.go#L432

plugins/dns/plugin.go
```diff
@@ -18,12 +18,22 @@ func init() {
...
 var (
-	dnsServers []net.IP
+	dnsServers6 []net.IP
+	dnsServers4 []net.IP
 )
 
 func setupDNS6(args ...string) (handler.Handler6, error) {
-	// TODO setup function for IPv6
-	log.Warning("not implemented for IPv6")
+	if len(args) < 1 {
```
pmazzini: How about having a more generic `setupDNS(v6 bool, args ...string)` to avoid code dup. Similar to:
https://github.com/coredhcp/coredhcp/blob/master/plugins/file/plugin.go#L197

pmazzini:
Thanks for the PR!

plugins/dns/plugin.go
Natolumin: Looks like this:
```go
func setupDNS(v int, args ...string) (handler.Handler4, handler.Handler6, error) {
	log.Printf("loaded plugin for DHCPv%d.", v)
	if len(args) < 1 {
		return nil, nil, errors.New("need at least one DNS server")
	}
	for _, arg := range args {
		DNSServer := net.ParseIP(arg)
		if v == 4 {
			if DNSServer.To4() == nil {
				return Handler4, nil, errors.New("expected an DNS server address, got: " + arg)
			}
			dnsServers4 = append(dnsServers4, DNSServer)
			log.Infof("loaded %d DNS servers.", len(dnsServers4))
		} else {
			if DNSServer.To16() == nil {
				return nil, Handler6, errors.New("expected an DNS server address, got: " + arg)
			}
			dnsServers6 = append(dnsServers6, DNSServer)
			log.Infof("loaded %d DNS servers.", len(dnsServers6))
		}
	}
	return Handler4, Handler6, nil
}
```

To me it doesn't really look better (code duplication is just moved into ifs instead of at the function level mostly), but I'm not attached to one or the other; if you think this is better I'll put it in.

plugins/dns/plugin.go
Natolumin: Not exactly. There's a difference between : 

- There is not ParameterRequest option in the message
- There is a ParameterRequest option, which does not contain OptionDomainNameServer

IsOptionRequested returns the same (false) in both cases, while this only matches the second case on purpose (see the comments right below and in the else block)

plugins/dns/plugin.go
Natolumin: Same as below, this doesn't differentiate "Option is there but doesn't contain the code" and "Option is not there". For v6 this matters less though, since clients MUST include the option; I guess we could act the same in both cases, but I think we should drop malformed packets as much as possible to avoid bad surprises later

plugins/dns/plugin.go
pmazzini: > To me it doesn't really look better

Ok, let's leave it like this for now then.

plugins/dns/plugin.go
pmazzini: How about adding the option only when explicitly requested? Why is it being added when there is no `ParameterRequest` option present?

plugins/dns/plugin.go
pmazzini: > but I think we should drop malformed packets as much as possible to avoid bad surprises later

I agree but this doesn't seem to me the proper place where to do packet validation. Should we move this validation somewhere in the library? Maybe in `NewRequestFromAdvertise` and `NewReplyFromMessage`.

plugins/dns/plugin.go
Natolumin: I based this decision on https://tools.ietf.org/html/rfc2131#section-3.5, specifically: 

> Not all clients require initialization of all parameters listed in
> Appendix A.  Two techniques are used to reduce the number of
> parameters transmitted from the server to the client. [...]  Second, in its
> initial DHCPDISCOVER or DHCPREQUEST message, a client may provide the
> server with a list of specific parameters the client is interested in.

Which I interpret, especially the "reduce the number" and "may provide", as saying that if there is no option, the server should send all non-default parameters

I agree that it's not especially clear and there is no MUST or SHOULD to give a definitive answer. But compared to DHCPv6 (where OptionRequestOption is MUST in SOLICIT/REQUEST/etc) it seems the intent is different

https://tools.ietf.org/html/rfc2131#section-1.3 has similar language to say the same thing slightly differently: 

>   Not all of these parameters are required for a newly initialized
>   client.  A client and server may negotiate for the transmission of
>   only those parameters required by the client or specific to a
>   particular subnet.

plugins/dns/plugin.go
Natolumin: Right, it's probably a good idea to put validation in the library indeed.
I think it's better not to overload simple functions like NewYFromX which can be used to create useful synthetic messages for testing even if they're invalid, but we can discuss that in an issue in the library

I'll remove this check here and move to IsOptionRequested

plugins/dns/plugin.go
Natolumin: Maybe a solution here is to reverse the logic of IsOptionRequested in the library, to return true for everything if there is no ParameterRequest, instead of false ?

pmazzini: > Maybe a solution here is to reverse the logic of IsOptionRequested in the library, to return true for everything if there is no ParameterRequest, instead of false ?

Ok, sounds good to me.

Natolumin:
out of curiosity: is it policy to squash all PRs for merging ? I made separate commits for a reason (especially easier revert and more helpful history), so it's a bit disappointing

pmazzini:
There is no policy, I've been squashing them because not all PRs have the separate commits as clean as yours. I'll take it into consideration for the following PRs.

insomniacslk:
hey @Natolumin , thanks for the contribution and welcome back! Generally we squash related commits into one, except when the author explicitly wants to keep them separate (or if we notice that they should stay separate). I guess it was just an oversight in this case

Issue #52, opened 2019-09-13
Listening on both multicast and unicast doesn't work (dhcpv6)


Natolumin:
in #28 and https://github.com/insomniacslk/dhcp/pull/278, connection handling and joining multicast groups was refactored. However, it's not possible (at least in linux) to listen on a non-wildcard unicast address, *and* join multicast groups. If you try now in coredhcp, you'll see the setsockopt call in strace but no traffic received on the multicast address, if there is a `listen` stanza with a unicast address in config.yml. So this: 
```yaml
server6:
    interface: eth0
    listen: 2001:db8:a::2
```
*Only* receives packets for the unicast address and **not** for `ff02::1:2%eth0`, even though the group is joined successfully: (from strace log:)
```
setsockopt(3, SOL_IPV6, MCAST_JOIN_GROUP, {gr_interface=if_nametoindex("eth0"), gr_group={sa_family=AF_INET6, sin6_port=htons(0), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "ff02::1:2", &sin6_addr), sin6_scope_id=0}}, 136)         = 0
```

Left column describes what you listen on, header describes how many multicast groups you join, table describes what traffic you end up receiving

Join > Listen v  | None | One | Multiple
----------------|--------|-------|-----------
**Unicast** | Unicast only | Unicast only | Unicast only
**Multicast** | None | Multicast if it matches the listening address | Multicast, only groups matching the listening address
**Wildcard** | Unicast only | Unicast + Multicast | Unicast + Multicast

Notably, there is no way to listen on a specific unicast address (not wildcard) and receive multicast traffic on the same socket.

I would fix this by implementing a more generic `listen` that can receive multiple addresses:
 * By default, it should listen to both RFC-defined multicast addresses (Ideally it should *not* accept unicast in this mode, but go doesn't let you do that and forces you to bind on `[::]` if you want any multicast, see [1], so we're forced to always bind to wildcard)
 * Adding one (or more) addresses in `listen` opens *separate* sockets to listen on the additional addresses, unicast or multicast without distinction
 * A separate option can be used to disable default multicast addresses (`multicast: no` ?)
 * Remove the `interface` configuration option, to fold it into the specific `listen` stanzas using the standard `addr%iface` notation (mandatory for link-local, recommended for multicast, not necessary for global unicast)

Example:
```yaml
server6: 
    listen: 2001:db8::1 fe80::a%eth0 ff08::1:3%eth1
[...]

Issue #50, opened 2019-09-12
Handling relay-forwarded messages is error-prone


Natolumin:
in #49 I had to fix some logic in 2 plugins for handling relayed messages. Since the full request message is passed to the plugin handlers, in most cases they have to unwind the relay-forward stack of encapsulated messages to get to the options of the inner message. This is error prone because most people won't think to test in relayed contexts (esp. since creating a test harness for relayed contexts is kinda complex).

Passing the inner message to the handlers is probably a bad idea since it removes some valuable information (eg `ExtractMac` is much less powerful) and outright prevents some types of plugins from being written (eg choosing an ip range based on the network segment of the client)
Even in the current scenario, the link-address and peer-address for the *current* server are unavailable to plugins (preventing, for example, writing a relay plugin)

I have a few ideas to make this less dangerous and more useful, but there are probably others: 

* Change the handlers to accept a more generic context interface than just the request message. For example this could expose the inner message directly, and the full relay chain separately, while being (more) easily extensible in the future by adding functions to the interface. This requires changing all the current handlers, but there aren't that many. rough example:
```go
// Request transmits context information on the request received by the server
type Request interface{
	// InnerMessage is the request message received, stripped of all relay-forw or relay-repl information
	InnerMessage() *dhcpv6.Message
	// ForwardChain is the list of relays the message passed through
	ForwardChain() []struct{PeerAddress net.IP; LinkAddress net.IP}

	// PeerAddr returns the address of the client or relay agent from which the message was received
	PeerAddr() net.IP
}
```

* Pass the request message encapsulated in a synthetic relay-forward message, as if the current server relayed the message to itself. This makes the peer-address and link-address available to plugins, and ensures all plugins always see only relayed messages, so that any kind of testing at all will show the issue if decapsulation is not handled.

PR #58, opened 2019-09-18
Allow listening on multiple addresses 

Natolumin:
This implements the changes discussed in #52 

There is an additional remaining issue mentioned in https://github.com/coredhcp/coredhcp/issues/52#issuecomment-532621844 where, when listening on a socket that is not bound to an interface, reply packets, especially broadcast or link-local, may be sent on the wrong interface. 

I believe this is fixable as mentioned in the issue, so this is a draft until we decide on a solution (or decide this is fine for the moment and to commit this)

Individual commit messages have more information

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -111,59 +111,77 @@ func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
 	return plugins, nil
 }
 
-func (c *Config) getInterface(ver protocolVersion) (string, error) {
-	if err := protoVersionCheck(ver); err != nil {
-		return "", err
+// BUG(Natolumin): listen specifications of the form `[ip6]%iface:port` or
+// `[ip6]%iface` are not supported, even though they are the default format of
+// the `ss` utility in linux. Use `[ip6%iface]:port` instead
+
+// splitHostPort splits an address of the form ip%zone:port into ip,zone and port.
+// It still returns if any of these are unset (unlike net.SplitHostPort which
+// returns an error if there is no port)
+func splitHostPort(hostport string) (ip string, zone string, port string, err error) {
+	ip, port, err = net.SplitHostPort(hostport)
+	if err != nil {
+		// Either there is no port, or a more serious error.
+		// Supply a synthetic port to differentiate cases
+		var altErr error
+		if ip, _, altErr = net.SplitHostPort(hostport + ":0"); altErr != nil {
+			// Invalid even with a fake port. Return the original error
+			return
+		}
+		err = nil
 	}
-	directive := fmt.Sprintf("server%d.interface", ver)
-	if exists := c.v.Get(directive); exists == nil {
-		return "", ConfigErrorFromString("dhcpv%d: missing `%s` directive", ver, directive)
+	if i := strings.LastIndexByte(ip, '%'); i >= 0 {
+		ip, zone = ip[:i], ip[i+1:]
 	}
-	ifname := c.v.GetString(directive)
-	if ifname == "" {
-		return "", ConfigErrorFromString("dhcpv%d: missing `%s` directive", ver, directive)
-	}
-	return ifname, nil
+	return
 }
 
 func (c *Config) getListenAddress(ver protocolVersion) (*net.UDPAddr, error) {
 	if err := protoVersionCheck(ver); err != nil {
 		return nil, err
 	}
+
 	addr := c.v.GetString(fmt.Sprintf("server%d.listen", ver))
-	if addr == "" {
-		// return default listener
-		if ver == protocolV6 {
-			return &net.UDPAddr{
-				IP:   net.IPv6unspecified,
-				Port: dhcpv6.DefaultServerPort,
-			}, nil
-		}
-		return &net.UDPAddr{
-			IP:   net.IPv4zero,
-			Port: dhcpv4.ServerPort,
-		}, nil
-	}
-	ipStr, portStr, err := net.SplitHostPort(addr)
+	ipStr, ifname, portStr, err := splitHostPort(addr)
 	if err != nil {
 		return nil, ConfigErrorFromString("dhcpv%d: %v", ver, err)
 	}
+
 	ip := net.ParseIP(ipStr)
+	if ipStr == "" {
+		switch ver {
+		case protocolV4:
+			ip = net.IPv4zero
+		case protocolV6:
+			ip = net.IPv6unspecified
+		}
+	}
 	if ip == nil {
-		return nil, ConfigErrorFromString("dhcpv%d: invalid IP address in `listen` directive", ver)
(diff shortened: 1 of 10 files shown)


pmazzini:
Sorry for the delay. I am still evaluating this and #52. I would love to know @insomniacslk's opinion.

With this change we may need to pass the more information to the plugins. Each plugin may have a different behaviour depending on which interface it is going to reply.

Just out of curiosity, are you aiming to use this in a relay or a server?

Natolumin:
> With this change we may need to pass more information to the plugins. Each plugin may have a different behaviour depending on which interface it is going to reply.

So first, I want to clarify that this does not change behaviour if it is bound to an interface (like is mandatory today). If we merged this as-is and people use it the same way (always specifying an interface), it would work the same.

For where to put things (and I guess you mean the ideas with WriteMsgUDP I mentioned in #52)
I was intending to put the decision for the return path in MainHandler instead of delegating more to plugins. Or rather, the idea was to break the separation where the library has Serve() with a handler API, and coredhcp has MainHandler that does impedance adaptation between the Serve() handler API and the handler API of the coredhcp plugins. Basically merge Serve() and MainHandler(), and put them in coredhcp together, based on my belief that every downstream user will want a different Serve() and there's no reason to have it in the library. As a simple example, dhcplb does exactly that, handling its own connections and doesn't use the library's Serve()

But, you're right (and this is more related to #50 ), I do want to change the plugin API to pass more information, so we *can*, if we want, build plugins that act different depending on the return path, or other additional info that they don't have atm.

> Just out of curiosity, are you aiming to use this in a relay or a server?

Both. I want to make a relay plugin for coredhcp, so that we may relay some queries but still answer others directly. That said, for network-related questions, a directly-attached server (receiving and responding to clients without going through a relay) seems closer in behaviour and problems encountered to a relay than to a server receiving/responding through a relay anyway,

Of course if I've forgotten to explain some things or if the details are unclear I can try to explain better and/or with more detail.

Natolumin:
Well this turned into a whole bunch of stuff. I think this will be much easier to review going commit-per-commit rather than looking at the whole. If anything is unclear, please say so, I'll both answer and add documentation as needed

I tested running as a server with multiple interfaces, with both relayed and non-relayed clients. I think we'll soon need some kind of e2e testing harness if we want to make sure to avoid regressions

pmazzini:
I had an offline conversation with @insomniacslk and he expressed concerns with this change. He told me he will try to take a look at it. I'd like everybody on the same page before merging it.

Natolumin:
Of course ! There's a lot of it and probably some things to discuss - I'm in #coredhcp in the gophers slack and on freenode if either of you want to discuss this more interactively. This is only one possible implementation, I also want consensus before going through :)

config/config.go
```diff
@@ -106,69 +106,86 @@ func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
...
 	}
+
 	ip := net.ParseIP(ipStr)
+	if ipStr == "" {
+		switch ver {
+		case protocolV4:
+			ip = net.IPv4zero
+		case protocolV6:
+			ip = net.IPv6unspecified
+		}
```
insomniacslk: Should we have a default here?

config/config.go
```diff
@@ -106,69 +106,86 @@ func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
...
+	if ip4 := ip.To4(); (ver == protocolV6 && ip4 != nil) || (ver == protocolV4 && ip4 == nil) {
+		return nil, ConfigErrorFromString("dhcpv%d: not a valid IPv%d address in `listen` directive", ver, ver)
 	}
-	port, err := strconv.Atoi(portStr)
-	if err != nil {
-		return nil, ConfigErrorFromString("dhcpv%d: invalid `listen` port", ver)
+
+	var port int
+	if portStr == "" {
+		switch ver {
```
insomniacslk: Same here about default

config/config.go
```diff
@@ -106,69 +106,86 @@ func parsePlugins(pluginList []interface{}) ([]*PluginConfig, error) {
...
+		case protocolV4:
+			port = dhcpv4.ServerPort
+		case protocolV6:
+			port = dhcpv6.DefaultServerPort
+		}
+	} else {
+		port, err = strconv.Atoi(portStr)
+		if err != nil {
+			return nil, ConfigErrorFromString("dhcpv%d: invalid `listen` port", ver)
```
insomniacslk: Also reported the value of `err`

config/config.go
Natolumin: Since we're inside config/ and the `ver` variable we're switching over never comes from user input, taking the default would always indicate a bug. Even then, since `ipStr == ""`, `ParseIP` will have returned nil and we enter the `if ip == nil` right after this and error out anyway.

So not having a default is not an error; however we can decide it's better for clarity and add one. I don't think it's necessary, but I'm not opposed to it so it's your call

config/config.go
Natolumin: Same as above, taking the default would be a bug (`ver` doesn't come from user input), we can signal explicitly on that. 
This one is a bit different as it would result in port 0 being silently used instead of an error which is a bigger issue; I can add it

config/config.go
insomniacslk: yeah, it would indicate a bug. I am OK with either calling a `panic()` here, and be loud about the bug, or silently going ahead given how unlikely this case is

config/config.go
insomniacslk: yeah, and as above I'd `panic()` here, to be loud about the bug

config/config.go
```diff
@@ -41,9 +42,8 @@ func New() *Config {
 // ServerConfig holds a server configuration that is specific to either the
 // DHCPv6 server or the DHCPv4 server.
 type ServerConfig struct {
-	Listener  *net.UDPAddr
-	Interface string
```
insomniacslk: as we were discussing on Slack, we may want to retain the interface in the config file. Since we don't have a final proposal for the new config format yet, shall we leave it here for now?

config/config.go
```diff
@@ -220,3 +229,94 @@ func (c *Config) parseConfig(ver protocolVersion) error {
...
+		if err != nil {
+			return nil, err
+		}
+		l = append(l,
+			net.UDPAddr{IP: dhcpv6.AllDHCPServers, Port: dhcpv6.DefaultServerPort},
+			// XXX: Do we want to listen on [::] as default ?
+		)
+		return l, nil
+	}
+	return nil, errors.New("defaultListen: Incorret protocl version")
```
insomniacslk: typo "protocl"

config/config.go
Natolumin: I've re-added a shim to understand the interface keyword in the configuration file, to mean the same thing as before

MalteJ:
how can a plugin detect via which interface a message came in?

PR #64, opened 2019-11-15
Proposed redis plugin

kworm83:
This is a proposed plugin to get static lease information from redis for review.  Currently only supports IPv4 as I don't have much v6 equipment to test with.

### DIFF
--- a/plugins/redis/plugin.go
+++ b/plugins/redis/plugin.go
@@ -23,7 +23,6 @@ import (
 var (
 	log       = logger.GetLogger("plugins/redis")
 	pool      *redis.Pool
-	leaseTime time.Duration
 )
 
 func init() {
@@ -64,9 +63,6 @@ func Handler4(req, resp *dhcpv4.DHCPv4) (*dhcpv4.DHCPv4, bool) {
 		log.Printf("MAC %s has no ipv4 field...dropping request", req.ClientHWAddr.String())
 		return nil, true
 	}
-
-	// Set default lease time - may be overriden by hash field option
-	resp.Options.Update(dhcpv4.OptIPAddressLeaseTime(leaseTime))
 
 	// Loop through options returned and assign as needed
 	for option, value := range options {
@@ -133,16 +129,11 @@ func setupRedis4(args ...string) (handler.Handler4, error) {
 }
 
 func setupRedis(v6 bool, args ...string) (handler.Handler6, handler.Handler4, error) {
-	var err error
-	if len(args) < 2 {
-		return nil, nil, fmt.Errorf("invalid number of arguments, want: 2 (redis server:port, lease time), got: %d", len(args))
+	if len(args) < 1 {
+		return nil, nil, fmt.Errorf("invalid number of arguments, want: 1 (redis server:port), got: %d", len(args))
 	}
 	if args[0] == "" {
 		return nil, nil, errors.New("Redis server can't be empty")
-	}
-	leaseTime, err = time.ParseDuration(args[1])
-	if err != nil {
-		return Handler6, Handler4, fmt.Errorf("invalid duration: %v", args[1])
 	}
 
 	if v6 {
--- a/plugins/redis/plugin.go
+++ b/plugins/redis/plugin.go
@@ -20,9 +20,11 @@ import (
 )
 
 // various global variables
-var log = logger.GetLogger("plugins/redis")
-var pool *redis.Pool
-var leaseTime time.Duration
+var (
+	log = logger.GetLogger("plugins/redis")
+	pool *redis.Pool
+	leaseTime time.Duration
+)
 
 func init() {
 	plugins.RegisterPlugin("redis", setupRedis6, setupRedis4)
@@ -39,17 +41,32 @@ func Handler4(req, resp *dhcpv4.DHCPv4) (*dhcpv4.DHCPv4, bool) {
 	// Get redis connection from pool
 	conn := pool.Get()
 
+	// defer redis connection close so we don't leak connections
 	defer conn.Close()
 
 	// Get all options for a MAC
 	options, err := redis.StringMap(conn.Do("HGETALL", "mac:"+req.ClientHWAddr.String()))
-	if err == redis.ErrNil || options["ipv4"] == "" {
-		log.Printf("MAC address %s is not allowed", req.ClientHWAddr.String())
-		return nil, true
-	} else if err != nil {
-		log.Printf("Redis error: %s", err)
+
+	// Handle redis error
+	if err != nil {
+				log.Printf("Redis error: %s...dropping request", err)
+			return nil, true
+		}
+	
+	// Handle no hash found
+	if len(options) == 0  {
+		log.Printf("MAC %s not found...dropping request", req.ClientHWAddr.String())
 		return nil, true
 	}
+
+	// Handle no ipv4 field
+	if options["ipv4"] == "" {
+		log.Printf("MAC %s has no ipv4 field...dropping request", req.ClientHWAddr.String())
+		return nil, true
+	}
+
+	// Set default lease time - may be overriden by hash field option
+	resp.Options.Update(dhcpv4.OptIPAddressLeaseTime(leaseTime))
 
 	// Loop through options returned and assign as needed
 	for option, value := range options {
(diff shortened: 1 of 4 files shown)


pmazzini:
Thanks for your contribution @kworm83!

* Please sign the commit:

```
git commit --amend --signoff
git push --force-with-lease origin redis-plugin
```

* I am worried about the new dependency introduced by the plugin. It will be a dependency even if the plugin is not used. cc @insomniacslk Thoughts?

plugins/redis/plugin.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/coredhcp/coredhcp/plugins"
+	"github.com/gomodule/redigo/redis"
+	"github.com/insomniacslk/dhcp/dhcpv4"
+	"github.com/insomniacslk/dhcp/dhcpv6"
+)
+
+// various global variables
+var log = logger.GetLogger("plugins/redis")
+var pool *redis.Pool
+var LeaseTime time.Duration
```
pmazzini: Does this need to be exported? If not start it with lower case.

kworm83: Oops...I have changed it to lowercase.

kworm83:
@pmazzini that is a good point about the redigo dependency.  I have to admit I am not familiar enough with the go build/module system to know if there is a way to have a conditional dependency like a #define in c.

insomniacslk:
> I am worried about the new dependency introduced by the plugin. It will be a dependency even if the plugin is not used. cc @insomniacslk Thoughts?

As long as it doesn't introduce a CGo dependency I'm not against adding new modules. However, importing it in main go means that it will be compiled and shipped even to who doesn't want to use redis, which is not a core feature, rather a nice-to-have for some users.

This problem will rise again, and we should resolve it for everyone. I'm thinking that we could split the core framework and plugins from the  other plugins, by using another subrepo, e.g. coredhcp/extra or something on those lines.

This means what we would have a default main.go with just core plugins, and the users would have to customize it. Some other projects have a builder that lets people specify the plugins to compile in (and under the hood generates a custom main.go), for example Caddy. Some others like CoreDNS have a script to generate it IIRC

insomniacslk:
On an unrelated note we should also document the status of each plugin, and whether they implement v6, v4 or both

kworm83:
Personally I like your idea of a sub-repo and users just editing the main.go to add the "extra" modules they want.

pmazzini:
> As long as it doesn't introduce a CGo dependency I'm not against adding new modules.

> Personally I like your idea of a sub-repo

Ok. Let move forward and if plugins continue to grow we can move the non core ones to a different repo.

pmazzini:
There are some merge conflicts.

plugins/redis/plugin.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/coredhcp/coredhcp/plugins"
+	"github.com/gomodule/redigo/redis"
+	"github.com/insomniacslk/dhcp/dhcpv4"
+	"github.com/insomniacslk/dhcp/dhcpv6"
+)
+
+// various global variables
+var log = logger.GetLogger("plugins/redis")
+var pool *redis.Pool
+var leaseTime time.Duration
```
pmazzini: NIT: Better to have one `var`:

```
var (
  log = 
  pool =
)
```

plugins/redis/plugin.go
```diff
@@ -0,0 +1,147 @@
...
+// Handler4 handles DHCPv4 packets for the redis plugin
+func Handler4(req, resp *dhcpv4.DHCPv4) (*dhcpv4.DHCPv4, bool) {
+	// Get redis connection from pool
+	conn := pool.Get()
+
+	defer conn.Close()
+
+	// Get all options for a MAC
+	options, err := redis.StringMap(conn.Do("HGETALL", "mac:"+req.ClientHWAddr.String()))
+	if err == redis.ErrNil || options["ipv4"] == "" {
```
pmazzini: When is this `redis.ErrNil` returned?

plugins/redis/plugin.go
```diff
@@ -0,0 +1,147 @@
...
+
+		default:
+			log.Printf("found unhandled option %s for MAC %s", option, req.ClientHWAddr.String())
+		}
+	}
+
+	// Set lease time
+	resp.Options.Update(dhcpv4.OptIPAddressLeaseTime(leaseTime))
+
+	// If request has Relay Agent Info copy it to the reply
```
pmazzini: This is not specific to the redis plugin, I think it should be moved to the main handler:
https://github.com/coredhcp/coredhcp/blob/master/coredhcp.go#L169

plugins/redis/plugin.go
kworm83: I will fix this

plugins/redis/plugin.go
kworm83: It would be returned when there are no values found for the key.  So basically that is saying if no values are found or there is not an ipv4 address field.

plugins/redis/plugin.go
kworm83: I was actually thinking that I should add support for a redis field for lease time and if it exists then update it.   Perhaps there could be a default in the main handler and then the redis plugin over-ride it.  What do you think about that?

kworm83:
On the merge conflicts I needed the updated insomniacslk/dhcp version to support the relay agent options.

plugins/redis/plugin.go
pmazzini: I was referring to the Relay Agent Info not the lease time. I wouldn't place the lease time in the main handler. Maybe we can have a lease time plugin and the redis plugin over-ride it.

plugins/redis/plugin.go
pmazzini: Ok, I would add a comment to clarify the `redis.ErrNil` error.

plugins/redis/plugin.go
kworm83: ahhh...sorry...yes the relay agent info is definitely not specific and to be honest I'm not 100% sure it has to be copied according to the RFCs but from wireshark captures the ISC servers copy it.  For the lease time I think the current way of the plugin having a lease time option set in the config file works fine and I could add it to allow redis to override in that plugin.  Do you think I should add support for more options in the redis plugin or just add them when/if anyone else needed them?

kworm83: Do you want me to make an attempt at moving the Relay Agent Info stuff into the main handler in this pull request?

Natolumin: I think it should go either into the dhcp library instead (github.com/insomniacslk/dhcp), to be included in `dhcpv4.NewReplyFromRequest`; Or in a separate plugin.

The RFC says "DHCP servers claiming to support the Relay Agent Information option SHALL echo the entire contents of the Relay Agent Information option in all replies", but we don't have to support it, so it's fine to just drop it from this

kworm83: It seems like adding another plugin would not be the best solution if it should be copied if it exists...that could potentially lead to unexpected behavior for a new user switch from another server and missing that they need to enable a plugin.  The dhcp library NewReplyFromRequest seems to be the most logical place for it.  What do you think @insomniacslk ?

pmazzini: Yes, maybe the library is a better place for it. I have seen some relays dropping the packet if the option in not there.

pmazzini:
> On the merge conflicts I needed the updated insomniacslk/dhcp version to support the relay agent options.

That is fine, can you rebase on top of master?

plugins/redis/plugin.go
pmazzini: It definitely should not be added here. We accept a library PR :) Otherwise, I can add it.

kworm83: I can take a stab at the library PR later today.

kworm83:
I fixed the variable declaration nit in redis/plugin.go and cleaned up the format of the log messages.  I also re-did redis error checking to differentiate between a hash key not being found vs not having the ipv4 field.  Also removed the relay options copy from the plugin.

plugins/redis/plugin.go
```diff
@@ -0,0 +1,164 @@
...
+
+func setupRedis(v6 bool, args ...string) (handler.Handler6, handler.Handler4, error) {
+	var err error
+	if len(args) < 2 {
+		return nil, nil, fmt.Errorf("invalid number of arguments, want: 2 (redis server:port, lease time), got: %d", len(args))
+	}
+	if args[0] == "" {
+		return nil, nil, errors.New("Redis server can't be empty")
+	}
+	leaseTime, err = time.ParseDuration(args[1])
```
pmazzini: How about moving the default lease time to a separate plugin?

pmazzini: Other than this, it looks good to me.

kworm83: I don't have a problem with that.  I will remove the default lease time from the redis plugin and make a new PR for a default lease time plugin.

kworm83:
So do I need to do anything else for it to be merged?

pmazzini:
I am letting other people comment on it. I'll merge it tomorrow if there are no comments.

PR #66, opened 2019-11-19
Initial commit of a proposed default lease time plugin.

kworm83:
This is an initial pass at a default lease time plugin as proposed by @pmazzini


plugins/lease_time/plugin.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+var log = logger.GetLogger("plugins/lease_time")
+
+func init() {
+	plugins.RegisterPlugin("lease_time", setupLeaseTime6, setupLeaseTime4)
+}
+
+var (
+	v6LeaseTime time.Duration
```
golangcibot: var `v6LeaseTime` is unused (from `unused`)

plugins/lease_time/plugin.go
```diff
@@ -0,0 +1,70 @@
...
```
golangcibot: var `v6LeaseTime` is unused (from `unused`)

kworm83: go build works fine but golangcibot is correct it is not currently used other than being setup from the config file.   I'm not familiar with v6 at all...anyone care to help out with what would need to be done to update the lease time on the v6 side?

Natolumin: There is no single concept of a "lease time" in v6. Each IA_* option has its own time values, and there are multiple (there's preferred and valid lifetimes, and then there are T1 and T2 times for when the client should resend a RENEW or a SOLICIT). I'd say drop all the v6 logic at this point, it doesn't translate directly at all

plugins/lease_time/plugin.go
```diff
@@ -0,0 +1,70 @@
...
+	return resp, false
+}
+
+// Handler4 handles DHCPv4 packets for the lease_time plugin.
+func Handler4(req, resp *dhcpv4.DHCPv4) (*dhcpv4.DHCPv4, bool) {
+	if req.OpCode != dhcpv4.OpcodeBootRequest {
+		log.Warningf("not a BootRequest, ignoring")
+		return resp, false
+	}
+	resp.Options.Update(dhcpv4.OptIPAddressLeaseTime(v4LeaseTime))
```
Natolumin: So this *overwrites* the option all the time ? What about previous plugins that have already made decisions on lease times ? 
If the option is already present, I think this plugin should not touch it. It's useful when the assignment plugin doesn't have a strong opinion on lease times (because it's stateless / reservation based, like `file` or `redis`), but if the assignment plugin handles leases and keeps track of their state we can't go overwriting the values outside of them

plugins/lease_time/plugin.go
Natolumin: Alternatively I guess, iterate over all the options in the packet, and when you find an IA_* option with undefined times (how do we know which are undefined and which are explicitly zero ? I don't have a good answer to this) update it with the value from the configuration

plugins/lease_time/plugin.go
kworm83: Yes it would overwrite all the time.  My thought was that it's behavior would be dependent on the order in which it was specified in the plugin list, ie if you put it first it's a default or if you put it after a plugin that sets lease time then it's an overwrite.  I can see utility in both.  Do you think I should make it something like "- lease_time: duration overwrite" with overwrite being a bool?

Natolumin: Relying on the ordering seems fine, although a bit surprising/dangerous, and needs to be well documented then.
However, I really can't see a situation where overwriting is a good idea. Do you have one ? We can always add the option to overwrite in the future when we're sure there's a usecase for it

pmazzini:
Please sign the commit :)
* git commit --amend --signoff
* git push --force-with-lease origin lease-time-plugin

plugins/lease_time/plugin.go
kworm83: The main reason I can think of you would ever want to overwrite is that you want to change every lease time to a short interval for maintenance purposes and not have to touch/understand every plugin.   Personally I think we should either have it overwrite and document it or I should add a flag for overwrite - that way users could get the most functionality out of it without someone possibly needing to make an overwrite_lease_time plugin down the road.  I'm fine with whatever you guys think though.

kworm83: For example when I place CoreDHCP into pre-production on some real life traffic I will first set the existing DHCP server to have a short like 10 minute lease time.  That way if there are issues it will only be a few minutes before the clients make a request that could (hopefully) fix the issue.  If I have my static leases all in redis with each one having it's own lease time then I would have to update each one to the shorter time and then change it back later after testing.  Again probably not a huge use case out there but nice to have the flexibility to handle it for those that might.

Natolumin: In this direction, it works. The problem is if you override the lease time advertised to the client with a longer lease, then redis will expire the lease before the client, and everything breaks.

Maybe only allow overriding when it lowers the lease time then ? That's at least safe

kworm83: I must be missing something...if the lease_time plugin is say first in the list of plugins and it sets the lease time to 1m and then redis gets the field to set it to 2m then the lease time would be 2 minutes.   If redis plugin is first and sets it to 1m and then the lease_time sets it to 2m then the lease time will be 2m.   Redis is not keeping any concept of state like lease expiration time because it's purely static.   I do see your point with another plugin that might keep track of expiration.   I will make another commit shortly with an idea for a solution.

kworm83: With my latest commit configuring the plugin with "- lease_time 300s" will only set the lease time if it hasn't already been set.  If you configure as "- lease_time 300s overwrite" then it will set it regardless.

Natolumin: > Redis is not keeping any concept of state like lease expiration time because it's purely static

In that kind of case, the redis plugin should not set an expiration time at all, because since it's all static, it doesn't care; then this plugin doesn't need to override anything. So for the pre-production usecase you mentioned before there's no need to overwrite. 
If however the lease expiration is stored in redis, then extending it causes problems since the value stored in redis is different from the one advertised to the client

kworm83: The lease expiration is not stored in redis so there will be no issue with that.  The redis plugin will set the lease time to the field named leasetime if it exists in the hash for a mac.  So in my use case, for example, lets say there are 5000 mac address hashes with a lease time of 30 minutes and 5000 with a lease time of 4 hours.  Now I want to do some testing/changing of the DHCP server, etc. so I want to temporarily set the lease time for everything to 5 minutes (and wait until the 4 hours are up so everything is requesting at 5 minute intervals).  The overwrite option would allow me to do that without having to update 10000 records in redis and change them back when I'm done.

Natolumin: OK, I get it now. I'm not sure I understand why the redis plugin is architected that way, but that's besides the point.

We'll need a way to support both this usecase and the stateful leases usecase in the future, but for now (at least for me) this looks reasonable.

kworm83: > I'm not sure I understand why the redis plugin is architected that way, but that's besides the >point.
Do you mean why the redis plugin supports allowing you to optionally set a lease time per mac address or why it doesn't keep an expiration time?

For the stateful use case the default for this lease_time plugin should work fine....it won't set it if the option already exists.

pmazzini:
There is a merge conflict. Can you rebase on top of master?

plugins/lease_time/plugin.go
```diff
@@ -0,0 +1,66 @@
...
+	leaseTime, err := time.ParseDuration(args[0])
+	if err != nil {
+		log.Warningf("invalid duration: %v", args[0])
+		return nil, errors.New("lease_time failed to initialize")
+	}
+	v4LeaseTime = leaseTime
+
+	// If overwrite is true then the lease time will be set regardless
+	// of if it has already been set by another plugin
+	overwrite = false
```
pmazzini: I am not a fan of this flag, I would simply document its behavior and consequences.

kworm83: I removed the flag and changed it to only set the lease time if the option does not already exist.  I think that will be less confusing to the larger audience that may use CoreDHCP.  I think @Natolumin would prefer that as well and he had a very good point about issues with overwriting the lease time with other plugins that may track expiration.

pmazzini:
The plugin will only be useful to set a default lease time when using static leases. I am ok with it. @Natolumin?

plugins/lease_time/plugin.go
```diff
@@ -0,0 +1,54 @@
...
+
+var (
+	log         = logger.GetLogger("plugins/lease_time")
+	v4LeaseTime time.Duration
+)
+
+// Handler4 handles DHCPv4 packets for the lease_time plugin.
+func Handler4(req, resp *dhcpv4.DHCPv4) (*dhcpv4.DHCPv4, bool) {
+	if req.OpCode != dhcpv4.OpcodeBootRequest {
+		log.Warningf("not a BootRequest, ignoring")
```
Natolumin: This should be an Infof at the max, probably a Debugf

plugins/lease_time/plugin.go
```diff
@@ -0,0 +1,54 @@
...
+	if !resp.Options.Has(dhcpv4.OptionIPAddressLeaseTime) {
+		resp.Options.Update(dhcpv4.OptIPAddressLeaseTime(v4LeaseTime))
+	}
+	return resp, false
+}
+
+func setupLeaseTime4(args ...string) (handler.Handler4, error) {
+	log.Print("loading `lease_time` plugin for DHCPv4")
+	if len(args) < 1 {
+		log.Warning("No default lease time provided")
```
Natolumin: This is an error, you're even returning one right on the next line

plugins/lease_time/plugin.go
```diff
@@ -0,0 +1,54 @@
...
+		return nil, errors.New("lease_time failed to initialize")
+	}
+
+	leaseTime, err := time.ParseDuration(args[0])
+	if err != nil {
+		log.Warningf("invalid duration: %v", args[0])
```
Natolumin: Same here, Errorf()

plugins/lease_time/plugin.go
kworm83: I removed this logging completely...not sure anyone would even want to debug that.

plugins/lease_time/plugin.go
kworm83: Fixed

plugins/lease_time/plugin.go
kworm83: Fixed

PR #67, opened 2019-11-21
Relay reply

insomniacslk:
If the request comes from an ephemeral source port, the server will
reply on such port. However the relay will be listening on server port
instead, so the reply is dropped. With this patch the reply always goes
to the server port if the request is a relay.

### DIFF
--- a/coredhcp.go
+++ b/coredhcp.go
@@ -160,7 +160,14 @@ func (s *Server) MainHandler6(conn net.PacketConn, peer net.Addr, req dhcpv6.DHC
 		resp = tmp
 	}
 
-	if _, err := conn.WriteTo(resp.ToBytes(), peer); err != nil {
+	var newPeer = peer
+	if req.IsRelay() {
+		newPeer = &net.UDPAddr{
+			IP:   peer.(*net.UDPAddr).IP,
+			Port: dhcpv6.DefaultServerPort,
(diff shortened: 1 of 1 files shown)


insomniacslk:
CC @pmazzini since you found the solution :P

coredhcp.go
```diff
@@ -160,7 +160,14 @@ func (s *Server) MainHandler6(conn net.PacketConn, peer net.Addr, req dhcpv6.DHC
 		resp = tmp
 	}
 
-	if _, err := conn.WriteTo(resp.ToBytes(), peer); err != nil {
+	var newPeer = peer
```
pmazzini: I think this will break non relay replies since we are nos specifying a response port for those.

pmazzini: Disregard

Natolumin:
This isn't compliant with the [Relevant RFC8357](https://tools.ietf.org/html/rfc8357) (emphasis mine): 
>Relay agents implementing this specification may be configured
   instead to 1) use a source port number other than 547 when relaying
   messages toward servers and 2) receive responses toward clients on
   **that same port.**

Supporting behaviour seen in the wild outside the RFC is imo fine, however this supports noncompliant behaviour by breaking compliant behaviour.

I think this must be reverted

PR #69, opened 2019-11-21
Revert "Relay reply"

Natolumin:
Reverts coredhcp/coredhcp#67

Copying the relevant portion of the RFC from https://github.com/coredhcp/coredhcp/pull/67#issuecomment-557078623: 
> Relay agents implementing this specification may be configured
instead to 1) use a source port number other than 547 when relaying
messages toward servers and 2) receive responses toward clients on
that same port.

Lower down: 
>Additionally, the IPv6 server MUST check and
   use the UDP source port from the UDP packet of the Relay-forward
   message in replying to the relay agent.

### DIFF
--- a/coredhcp.go
+++ b/coredhcp.go
@@ -160,14 +160,7 @@ func (s *Server) MainHandler6(conn net.PacketConn, peer net.Addr, req dhcpv6.DHC
 		resp = tmp
 	}
 
-	var newPeer = peer
-	if req.IsRelay() {
-		newPeer = &net.UDPAddr{
-			IP:   peer.(*net.UDPAddr).IP,
-			Port: dhcpv6.DefaultServerPort,
-		}
-	}
-	if _, err := conn.WriteTo(resp.ToBytes(), newPeer); err != nil {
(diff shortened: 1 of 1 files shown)


pmazzini:
Right, it makes sense, I didn't go through the RFC, thanks for bringing this up.

Sign the commit?

I think @insomniacslk want's to keep the behavior with an additional flag.

pmazzini:
Or we can simply go ahead with the revert.

pmazzini:
looks good after DCO.

insomniacslk:
Yeah i did this to comply with what dhcplb does, but as you noted the RFC states the opposite and the original behaviour is the right one. @pmazzini agreed to fix dhcplb

insomniacslk:
@Natolumin can you resubmit after amending with DCO?

Natolumin:
> Sign the commit?

Ah right, sorry, should have guessed the webUI wouldn't do it. 

> I think @insomniacslk want's to keep the behavior with an additional flag.

Out of curiosity, is there any chance the relay is fixable here ? Do you have an example packet ?

PR #71, opened 2019-11-21
Added CoreDHCP generator tool

insomniacslk:
This tool generates the source file for a CoreDHCP server based on the
specified plugins. The tool must be run from its source directory,
otherwise it requires the -template argument.
Example:
```
./coredhcp-generator \
        github.com/coredhcp/coredhcp/plugins/{dns,file}
```
will generate a file called coredhcp.go in a temporary directory.


.gitignore
```diff
@@ -4,4 +4,5 @@ cmds/coredhcp/leases4.txt
 cmds/coredhcp/config.yml
 cmds/coredhcp/coredhcp
 cmds/client/client
+cmds/coredhcp-generator/coredhcp-generator
```
pmazzini: I would simply call it generator or server generator, the coredhcp feels redundant.

cmds/coredhcp-generator/main.go
```diff
@@ -0,0 +1,64 @@
...
+		if err != nil {
+			log.Fatalf("Cannot create temporary directory: %v", err)
+		}
+		outfile = path.Join(tmpdir, "coredhcp.go")
+	}
+	log.Printf("Generating output file '%s' with %d plugins", outfile, len(plugins))
+	outFD, err := os.OpenFile(outfile, os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		log.Fatalf("Failed to create output file '%s': %v", outfile, err)
+	}
```
pmazzini: defer outFD.Close()

.gitignore
pmazzini: I don't have a strong opinion about it though.

insomniacslk: it was the original name, but if people do go install github.com/coredhcp/coredhcp/cmds/generator they will end up with a `generator` binary in their PATH. I used the longer name `coredhcp-generator` to remove ambiguity from the CLI name

cmds/coredhcp-generator/main.go
insomniacslk: good catch!

insomniacslk:
mh, something that this tool revealed is that we lose the modules support in the generated directory (it works if I copy the go.{sum,mod} by hand from the repo root but it's not a good solution).

I'm merging this anyway because it doesn't break existing workflows, but I'll address this soon

PR #73, opened 2019-11-22
[cmds/contest] Redis plugin is not compiled by default

insomniacslk:
Not every installation needs the Redis plugin, so we moved it out of
the core plugins, and this patch also stops compiling it by default,
reducing dependencies and binary size of the default cmd.
To compile the redis plugin and other plugins, use
cmds/coredhcp-generator.


insomniacslk:
CC @kworm83

PR #75, opened 2019-12-05
Setup network namespaces in travis

Natolumin:
This is an extremely WIP PR to demo a setup of network namespaces so we can run real complete client-server exchanges. (cc @insomniacslk)

There is an example test included. It doesn't pass, but it shows that the setup allows communication between client & server: 

`2019/12/05 10:42:18 Handling request from [fe80::f461:cbff:feae:a8ee%ens4]:546`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ script: |
   set -x
   case $TEST_SUITE in
   unit)
+      ./.travis/setup-integ.sh
       ./.travis/tests.sh
       ;;
   linters)
--- a/.travis/setup-integ.sh
+++ b/.travis/setup-integ.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+set -ex
+
+if [[ $UID -ne 0 ]] # TODO: check for permissions instead (we can have CAP_NET_ADMIN without root)
+then
+    sudo "$0" "$@"
+    exit $?
+fi
+
+# Topology: with one or 3 netns
+#
+# * 3-netns, for relay operations
+#  --------------------------------
+# | server (cdhcp_srv  <---------) | Upper netns
+#  -----------------------------|--
+#                               | (veth pair)
+#  -----------------------------|---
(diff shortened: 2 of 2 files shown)


Natolumin:
I haven't figured out yet how to get coverage of the entire project for e2e tests.
`go test -coverage` does not recurse into dependencies, so we only test the coverage of the e2e_test package itself.
 Maybe something like this: https://github.com/lawrencewoodman/roveralls ?

insomniacslk:
thanks @Natolumin ! I think we can split the netns setup and the actual integration tests in two parts, probably the most important work is the former. Right now I cannot reproduce as I'm on a Linux container on a Chromebook and the setup fails for me, but I'll retry on another machine shortly

insomniacslk:
if the current setup works, can you squash the commits into one for setup-integ.sh and another for the integ tests?

Natolumin:
I removed the integration test commit for now, I'll rework and add something either here if it happens soon, or later.

### micro documentation from the code 
maybe we should add that in a HACKING.md or something ? 
#### 3-netns, for when you have a relay
```
 --------------------------------
| server (cdhcp_srv  <---------) | Upper netns (coredhcp-upper)
 -----------------------------|--
                              | (veth pair)
 -----------------------------|--
| relay upper (cdhcp_relay_u <-) |
|                                | Relay netns (coredhcp-relay)
| relay lower (cdhcp_relay_d <-) |
 -----------------------------|--
                              | (veth pair)
------------------------------|--
|  client (cdhcp_cli <---------) | Lower netns (coredhcp-lower)
---------------------------------
```

#### 2-netns, for when your clients are directly connected
```
 --------------------------------
| server (cdhcp_srv  <---------) | Upper netns (coredhcp-direct-upper)
 -----------------------------|--
                              | (veth pair)
------------------------------|--
|  client (cdhcp_cli <---------) | Lower netns (coredhcp-direct-lower)
---------------------------------
```

PR #81, opened 2019-12-17
RFC/PoC: Support for dynamic plugin loading

Natolumin:
On linux and macOS, go supports building shared objects and loading them at runtime. This PR adds support in coredhcp for loading these, in the style of C/C++ dynamic plugins found in many server projects.

If this idea is ok, I'll work on some build tooling and additional documentation

## Instructions : 
* change the package name for the plugin to `main` (required to build as plugin)
* build the plugin (either in-tree or out-of-tree) with `go build -buildmode=plugin`. Rename it to `plugin_<name>.so`
* add the directory where your .so is in the config.yml as a top-level object:
  ```yaml
  dynamic_plugins: /home/coredhcp/plugins
  ```
  And load the plugin in one of your server statements
* **remove** the plugin from the `import` statement in `cmd/coredhcp/main.go` or in the file generated by coredhcp-generator
* run coredhcp, the plugin should load as if it were builtin

## Troubleshooting

You will most probably encounter the following error (don't forget to run with `-debug` until it works):
```
DEBUG coredhcp: Attempted to load plugin redis dynamically: could not load dynamic plugin redis: plugin.Open("coredhcp-plugins/plugin_redis"): plugin was built with a different version of package golang.org/x/sys/unix
```
Any package version mismatch will cause issues. At the moment both coredhcp/plugins and insomniacslk/dhcp do not use go modules, and building these libraries will use whatever version of dependencies you have in your gopath.
To help with this during local development, you can: 
1. run `go mod init` in both your repositories for `github.com/insomniacslk/dhcp` and `github.com/coredhcp/plugins`
1. add the following replace statements in the go.mod files:
  ```go
  // In coredhcp/coredhcp/go.mod
  replace github.com/insomniacslk/dhcp => <path to your checkout of insomniacslk/dhcp>
  // in coredhcp/plugins/go.mod
  replace (
          github.com/coredhcp/coredhcp => <path to your checkout of coredhcp>
          github.com/insomniacslk/dhcp => <path to your checkout of insomniacslk/dhcp>
  )
  ```

## Caveats
**This is not a replacement for coredhp-generator and fixing static build**
It is only supported: 
 * On linux and macOS
 * With cgo
 * When all included packages are the same versions (modules can somewhat help here)

It requires some additional tooling to build (changing packages for plugins to "main" for example)

### DIFF
--- a/cmds/coredhcp/config.yml.example
+++ b/cmds/coredhcp/config.yml.example
@@ -16,3 +16,5 @@ server4:
         # - router: 10.10.10.1
         # - netmask: 255.255.255.0
         # - range: leases.txt 10.10.10.100 10.10.10.200 60s
+
+dynamic_plugins: /usr/lib64/coredhcp/plugins
--- a/config/config.go
+++ b/config/config.go
@@ -5,8 +5,11 @@
 package config
 
 import (
+	"errors"
 	"fmt"
 	"net"
+	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -28,9 +31,10 @@ const (
 
 // Config holds the DHCPv6/v4 server configuration
 type Config struct {
-	v       *viper.Viper
-	Server6 *ServerConfig
-	Server4 *ServerConfig
+	v         *viper.Viper
+	Server6   *ServerConfig
+	Server4   *ServerConfig
+	PluginDir string
 }
 
 // New returns a new initialized instance of a Config object
@@ -63,6 +67,9 @@ func Load() (*Config, error) {
 	c.v.AddConfigPath("$HOME/.coredhcp/")
 	c.v.AddConfigPath("/etc/coredhcp/")
 	if err := c.v.ReadInConfig(); err != nil {
+		return nil, err
+	}
+	if err := c.parsePluginDir(); err != nil {
 		return nil, err
 	}
 	if err := c.parseConfig(protocolV6); err != nil {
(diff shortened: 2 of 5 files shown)


mdlayher:
As I understand it, Go plugins have historically not been well supported and in general it seems that many folks would prefer not to use them. Is this something CoreDNS (for example) is using? It seems like it's simpler and safer to me to configure and enable plugins once at startup.

DHCP clients work over an unreliable UDP connection anyway, so any minor server downtime would just cause the clients to back off and retry.

Natolumin:
> As I understand it, Go plugins have historically not been well supported and in general it seems that many folks would prefer not to use them. Is this something CoreDNS (for example) is using?

That's also my understanding, and this is why I emphatically don't want to replace the monolithic build with it. The documentation for plugin does describe it as supported at least, but yeah that doesn't mean much. 
It's an option, and I think it makes sense to have it in parallel because the code is straightforward, doesn't really interact with anything else, and there isn't much of it, so I don't expect difficulties arising from it.

>  It seems like it's simpler and safer to me to configure and enable plugins once at startup.

That is exactly what it does! All the loading happens at startup.
This is more of a crutch to help build a server from various plugins in a go world where even with modules dependency management is shit, and you need to build your own hacky scripts (see coredhcp-generator for our initial attempt, or the coredns builder system) to have any reasonable kind of out-of-tree plugin support.

> DHCP clients work over an unreliable UDP connection anyway, so any minor server downtime would just cause the clients to back off and retry.

I don't understand this comment, this PR doesn't change anything regarding this afaict ?

Issue #86, opened 2020-02-06
Plugin logging on registration can't be made silent


elmarco:
Hi,

When the plugins are init(), they log the registration:
```
[2020-02-06T14:47:46+01:00]  INFO plugins: Registering plugin "dns"
[2020-02-06T14:47:46+01:00]  INFO plugins: Registering plugin "file"
...
```

Unfortunately, I can't see a way to silent it, since init() is called early, before the program has the chance to change logrus behaviour. Please advise.

elmarco:
so my current workaround is to have a custom plugin that does:

```
func init() {
	log := logger.GetLogger("plugins")
	log.Logger.SetLevel(logrus.WarnLevel)
}
```

insomniacslk:
Hi @elmarco , the current method to register plugins is suboptimal, and will change. This is tracked in https://github.com/coredhcp/coredhcp/issues/45 and will let you do what you are asking here. It's faily simple to do, and I'll try to tackle it this week

insomniacslk:
for an example of how it will look like, see how we register plugins in another project, https://github.com/facebookincubator/contest/blob/master/cmds/contest/main.go#L90

insomniacslk:
@elmarco can you try with the latest commit? Now setting the log level globally should not print the plugin loading messages anymore.

We can also add a CLI flag for the purpose

elmarco:
works for me, thanks

PR #87, opened 2020-02-07
Register plugins explicitly

### LINKED ISSUE #45: Move plugin initialization to main.go
The plugin initialization should not happen in a "magic" `init()` call in the plugins. Rather, the plugins should expose a registration function that does the plugin set-up and returns plugin name and handlers. The framework should handle the rest.

insomniacslk:
Fixes #45
Do not use `init()` magic to register plugins, do it explicitly. This is
done by requiring plugins to declare a populated `Plugin` symbol of type
`plugins.Plugin`.


insomniacslk:
Related to #86

mdlayher:
LGTM, explicit is best.

PR #88, opened 2020-02-09
Added -loglevel flag

### LINKED ISSUE #86: Plugin logging on registration can't be made silent

insomniacslk:
And removed -debug flag. Now all the logging levels can be specified
with -loglevel. "none" is a special log level that disables logging
altogether.

Fixes #86 

### DIFF
--- a/cmds/coredhcp-generator/coredhcp.go.template
+++ b/cmds/coredhcp-generator/coredhcp.go.template
@@ -6,6 +6,8 @@ package main
 
 import (
 	"flag"
+	"fmt"
+	"io/ioutil"
 	"time"
 
 	"github.com/coredhcp/coredhcp"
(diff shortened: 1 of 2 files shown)


Natolumin:
-quiet usually means the opposite of -verbose, reducing the logging level
What's our long-term commitment for cli flags stability ? If we're allowed with changing the meaning in the future I'm ok with this, but otherwise this is essentially equivalent to `-logfile /dev/null`, so maybe not worth wasting a well-known flag name on if we can't change the meaning again

insomniacslk:
Fair point, what about `-loglevel [none,debug,info,warning,error]` ?

Natolumin:
That would be nice to have indeed

insomniacslk:
@Natolumin done

cmds/coredhcp-generator/coredhcp.go.template
```diff
@@ -21,9 +23,26 @@ import (
 var (
 	flagLogFile     = flag.String("logfile", "", "Name of the log file to append to. Default: stdout/stderr only")
 	flagLogNoStdout = flag.Bool("nostdout", false, "Disable logging to stdout/stderr")
-	flagDebug       = flag.Bool("debug", false, "Enable debug output")
+	flagLogLevel    = flag.String("loglevel", "info", fmt.Sprintf("Log level. One of %v", getLogLevels()))
 )
 
+var logLevels = map[string]func(*logrus.Logger){
+        "none":    func(l *logrus.Logger) { l.SetOutput(ioutil.Discard) },
```
Natolumin: Do we have a log level that avoids any output ?

With this, we'll emit (and format, etc) the log messages, then discard them. With a high level we can prevent the messages from even being evaluated

cmds/coredhcp/main.go
```diff
@@ -43,10 +62,14 @@ var desiredPlugins = []*plugins.Plugin{
...
-		log.Logger.SetLevel(logrus.DebugLevel)
-		log.Infof("Enabled debug logging")
+	fn, ok := logLevels[*flagLogLevel]
+	if !ok {
+		log.Fatalf("Invalid log level '%s'. Valid log levels are %v", *flagLogLevel, getLogLevels())
+	}
+	if *flagLogLevel != "none" {
+		log.Infof("Setting log level to '%s'", *flagLogLevel)
 	}
+	fn(log.Logger)
```
Natolumin: This should be first - if I pass loglevel critical, I don't want an info-level log to tell me I've set the loglevel to critical

cmds/coredhcp-generator/coredhcp.go.template
insomniacslk: not that I could find - the code would be more compact if it did...

cmds/coredhcp/main.go
insomniacslk: done

insomniacslk: I have also removed the "none" check which is not necessary anymore after the swap

insomniacslk:
thanks for the review, I have updated the PR according to the suggestions and will merge once tests are green

PR #90, opened 2020-02-13
serverid: fix v4ServerID copy

elmarco:
I am not very familiar with net.IP, but copying the first 4 bytes of
a v4 net.IP currently result in 0.0.0.0.

Signed-off-by: Marc-André Lureau <marcandre.lureau@redhat.com>

### DIFF
--- a/plugins/serverid/plugin.go
+++ b/plugins/serverid/plugin.go
@@ -110,7 +110,7 @@ func setup4(args ...string) (handler.Handler4, error) {
 	if serverID.To4() == nil {
 		return nil, errors.New("not a valid IPv4 address")
 	}
-	v4ServerID = serverID
+	v4ServerID = serverID.To4()
 	return Handler4, nil
 }
 


elmarco:
@Natolumin does that do the job?

```
diff --git a/plugins/serverid/plugin.go b/plugins/serverid/plugin.go
index cf4a665..82e712b 100644
--- a/plugins/serverid/plugin.go
+++ b/plugins/serverid/plugin.go
@@ -110,7 +110,7 @@ func setup4(args ...string) (handler.Handler4, error) {
        if serverID.To4() == nil {
                return nil, errors.New("not a valid IPv4 address")
        }
-       v4ServerID = serverID
+       v4ServerID = serverID.To4()
        return Handler4, nil
 }

```

Natolumin:
Yeah, that should be enough, it looks like that's the only place this value is set

plugins/serverid/plugin.go
```diff
@@ -92,8 +92,8 @@ func Handler4(req, resp *dhcpv4.DHCPv4) (*dhcpv4.DHCPv4, bool) {
 		log.Infof("requested server ID does not match this server's ID. Got %v, want %v", req.ServerIPAddr, v4ServerID)
 		return nil, true
 	}
-	resp.ServerIPAddr = make(net.IP, net.IPv4len)
-	copy(resp.ServerIPAddr[:], v4ServerID)
+	resp.ServerIPAddr = make(net.IP, len(v4ServerID))
```
pmazzini: Is this change needed after the change in line 113?

elmarco: probably not, but it looks more correct anyway

elmarco: probably not, but it looks more correct anyway

Natolumin: It doesn't look more correct to me: you have to know that somewhere else in the code, something ensures the address is 4 bytes.
If you also call To4() here in addition to the other place, it's more obvious that we're not putting an ipv6 address in an ipv4 variable

(This would be less of an issue if go had separate types for v4-only addresses, but it doesn't)

elmarco: it's more correct because that change actually fixes the bug too.
 I am not very familiar with go, but I don't get the [:] either.

pmazzini: I would stick to having `net.IPv4len` here and only change line 113.

elmarco: oh well, who cares

insomniacslk:
I'm late to the party but this change looks good after switching to `.To4()`, I'm merging it. Thanks everyone!

PR #98, opened 2020-04-21
Remove DHCPv6 stub implementations in unsupported plugins

Natolumin:
These plugins don't support IPv6. With the new explicit plugin
registration, completely omitting the setup function for ipv6 in that
case is now possible.


insomniacslk:
Thanks @Natolumin !

PR #99, opened 2020-04-23
Fix drift between coredhcp-generator/ and coredhcp/

Natolumin:
We have 2 entrypoints for coredhcp, the generator (for arbitrary plugin selection) and the "standard" cmds/coredhcp with only builtin plugins. They tend to drift from one another because they were not exactly the same and we don't think to update one when we change the other.

This PR does the following to fix this situation: 
 * Update the template in coredhcp-generator and main.go, so that they match. This fixes the breakage introduced by #58 which only updated cmds/coredhcp
 * Improve the generator so it can be scripted, and add a script to check that both entrypoints are identical
 * Add a script to travis to check this on each pr

There are some additional quality-of-life commits, see individual commit messages for more specific descriptions


cmds/coredhcp-generator/main.go
```diff
@@ -52,6 +54,13 @@ func main() {
 		if pl == "" {
 			continue
 		}
+		if !strings.ContainsRune(pl, '/') {
+			// A bare name was specified, not a full import path.
+			// Coredhcp plugins aren't in the standard library, assume this is one of the builtin plugins
```
insomniacslk, line 59: this assumption is not flawless: technically your import path may have package names without slashes (e.g. `import mymodule` would work if you have `$GOPATH/src/mymodule`). However this is very unlikely and I think we can run the risk. Maybe just update the comment to reflect this?

cmds/coredhcp/main.go
```diff
@@ -15,14 +15,15 @@ import (
...
 	"github.com/coredhcp/coredhcp/plugins"
-	"github.com/coredhcp/coredhcp/plugins/dns"
-	"github.com/coredhcp/coredhcp/plugins/file"
-	"github.com/coredhcp/coredhcp/plugins/leasetime"
-	"github.com/coredhcp/coredhcp/plugins/nbp"
-	"github.com/coredhcp/coredhcp/plugins/netmask"
-	rangepl "github.com/coredhcp/coredhcp/plugins/range"
-	"github.com/coredhcp/coredhcp/plugins/router"
-	"github.com/coredhcp/coredhcp/plugins/serverid"
+	pl_dns "github.com/coredhcp/coredhcp/plugins/dns"
```
insomniacslk, line 18: I'd like to have a comment here to explain why the `pl_` prefix is necessary, it would be confusing otherwise

.travis.yml
```diff
@@ -1,10 +1,12 @@
 language: go
 
-sudo: required
+os: linux
+dist: bionic
 
 go:
   - "1.12"
```
insomniacslk: I'd be ok with dropping 1.12 altogether too - 1.13 introduces new error handling (and I like `%w`) and so far we committed to supporting only the last two versions. Additionally we should switch to the "1.14.x" syntax so that Travis-CI can pick the latest minor version automatically

cmds/coredhcp-generator/main.go
Natolumin: Oh, I didn't consider that indeed. Since the `-from` version doesn't do this expansion, at least there is an easy workaround if anyone needs it; I'll update the comment and leave it at that

cmds/coredhcp/main.go
Natolumin: How about a "this is a generated file" header, like:
```
// This is a generated file, based on cmds/coredhcp-generator/coredhcp.go.template
// Regenerate with `coredhcp-generator -from core-modules.txt`
```
This doesn't directly address the plugin names but any weirdness will be explained by looking at the source template ?

.travis.yml
Natolumin: good idea on dropping 1.12, I'll do it in a separate PR as it needs a go.mod update as well

cmds/coredhcp/main.go
insomniacslk: yes, sounds good

PR #101, opened 2020-05-01
config: Allow overriding config location from flags

### LINKED ISSUE #93: dynamic config file location?
Slightly related to #78; but I think it would be great if the program is able to receive the config file location via flags, rather than preempting the location for the users (./config.yml, $HOME/.coredhcp/config.yml, /etc/coredhcp/config.yml).
For example, I should be able to start the app with `coredhcp -c /tmp/config.test.yaml`.
Not sure, but I think many users would benefit from this approach.

Natolumin:
This lets the user specify a configuration file to load, disregarding
the builtin heuristics for locating it

Fixes #93

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -54,14 +54,20 @@ type PluginConfig struct {
 
 // Load reads a configuration file and returns a Config object, or an error if
 // any.
-func Load() (*Config, error) {
+func Load(pathOverride string) (*Config, error) {
 	log.Print("Loading configuration")
 	c := New()
 	c.v.SetConfigType("yml")
-	c.v.SetConfigName("config")
-	c.v.AddConfigPath(".")
-	c.v.AddConfigPath("$HOME/.coredhcp/")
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -54,14 +54,20 @@ type PluginConfig struct {
...
-	c.v.SetConfigName("config")
-	c.v.AddConfigPath(".")
-	c.v.AddConfigPath("$HOME/.coredhcp/")
-	c.v.AddConfigPath("/etc/coredhcp/")
+	if pathOverride != "" {
+		c.v.SetConfigFile(pathOverride)
+	} else {
+		c.v.SetConfigName("config")
+		c.v.AddConfigPath(".")
+		c.v.AddConfigPath("$XDG_CONFIG_HOME/coredhcp/")
```
Natolumin, line 66: I've added this one, to follow XDG specs. I can remove it if we don't want to change this behavior now

PR #106, opened 2020-06-29
README: update example

### LINKED ISSUE #105: Listen on a specific interface
Is there a way to force the server to listen to a specific interface ? I'm not seeing anything about this in the README.MD and example files

insomniacslk:
Fixes #105

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,27 +12,21 @@ This is still a work-in-progress
 
 In CoreDHCP almost everything is implemented as a plugin. The order of plugins in the configuration matters: every request is evaluated calling each plugin in order, until one breaks the evaluation and responds to, or drops, the request.
 
-The following configuration runs a DHCPv6-only server, listening on all the interfaces, using a custom DUID-LL as server ID, and reading the leases from a text file.
+The following configuration runs a DHCPv6-only server, listening on all the interfaces, using a custom server ID and DNS, and reading the leases from a text file.
 
 ```
 server6:
-    listen: '[::]:547'
+    # this server will listen on all the available interfaces, on the default
+    # DHCPv6 server port, and will join the default multicast groups. For more
+    # control, see the `listen` directive in cmds/coredhcp/config.yml.example .
     plugins:
         - server_id: LL 00:de:ad:be:ef:00
         - file: "leases.txt"
-        # - dns: 8.8.8.8 8.8.4.4 2001:4860:4860::8888 2001:4860:4860::8844
-
-#server4:
-#    listen: '0.0.0.0:67'
-#    plugins:
-        # - server_id: 10.10.10.1
-        # - dns: 8.8.8.8 8.8.4.4
-        # - router: 10.10.10.1
-        # - netmask: 255.255.255.0
-        # - range: leases.txt 10.10.10.100 10.10.10.200 60s
+        - dns: 8.8.8.8 8.8.4.4 2001:4860:4860::8888 2001:4860:4860::8844
 ```
 
-See also [config.yml.example](cmds/coredhcp/config.yml.example).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,11 +16,12 @@ The following configuration runs a DHCPv6-only server, listening on all the inte
 
 ```
 server6:
-    listen: '[::]:547'
+    listen: '[::]:547'       # this listens on every interface
+    # - "[fe80::1%eth0]:547" # this will listen on eth0
```
Natolumin, line 20: Not sure if this makes it clear that `listen:` is normally a list, but also accepts a single address as a string if there is only one. Maybe the simplest is to only document the list version

README.md
```diff
@@ -16,11 +16,12 @@ The following configuration runs a DHCPv6-only server, listening on all the inte
...
     plugins:
-        - server_id: LL 00:de:ad:be:ef:00
-        - file: "leases.txt"
-        # - dns: 8.8.8.8 8.8.4.4 2001:4860:4860::8888 2001:4860:4860::8844
+        - server_id: LL 00:de:ad:be:ef:00  # set the server ID (option 2)
+        - file: "leases.txt"  # read static lease configuration from leases.txt
+        # - dns: 8.8.8.8 8.8.4.4 2001:4860:4860::8888 2001:4860:4860::8844  # set DNS (option 23)
 
 #server4:
 #    listen: '0.0.0.0:67'
```
Natolumin, line 27: Probably worth adding a v4 equivalent as well, eg `192.168.1.1%eth0`, or even `%eth0`

insomniacslk: I thought about it - but the v4 example is commented out. I'm thinking that this is not the right thing to do, and we should put in the readme the simplest example. The example config is already linked immediately after, and we should be probably clear that it's where people should look for for more information

insomniacslk: in other words I'm for reverting the README change and reinforce the wording to look at the example for more details

Natolumin: oh, yes, that would be simpler to manage and remember to update this way

README.md
```diff
@@ -12,27 +12,19 @@ This is still a work-in-progress
 
 In CoreDHCP almost everything is implemented as a plugin. The order of plugins in the configuration matters: every request is evaluated calling each plugin in order, until one breaks the evaluation and responds to, or drops, the request.
 
-The following configuration runs a DHCPv6-only server, listening on all the interfaces, using a custom DUID-LL as server ID, and reading the leases from a text file.
+The following configuration runs a DHCPv6-only server, listening on all the interfaces, using a custom server ID and DNS, and reading the leases from a text file.
 
 ```
 server6:
     listen: '[::]:547'
```
Natolumin, line 19: if we're going for minimum viable configuration, we can even just remove this line
This is even misleading because specifying the address manually will result in not actually joining the default multicast groups (whereas not having the `listen` at all will result in joining all_dhcp_servers and all_dhcp_servers_and_relays; but not listening on the unicast addresses)

I'll send a PR with better doc in the example config for the listen statement, I realize it's underdocumented

insomniacslk: How do we join the default multicast groups with an explicit line? I think we should have that

insomniacslk: Btw right now I'll remove the line, and add a comment that no `listen` directive will listen on all interfaces on the default multicast group (and we can expand on other syntaxes in the example config)

Natolumin: You just write them out:
```yaml
listen:
  - "[ff02::1:2]" # This will automatically expand to joining the group on each existing interface
  - "[ff05::1:3]"
```

insomniacslk: Updated again

insomniacslk: > You just write them out

Cool :) this should go to `config.yml.example`

PR #108, opened 2020-07-07
Prefix Delegation

Natolumin:
This implements a first pass for an IPv6 Prefix Delegation plugin, and happens to create some scaffolding for other fun things at the same time.

There is again a lot of stuff in here (sorry) so I'll try my best to provide explanations. Please ask if anything is unclear, I'll update the comments and/or readmes.

First 4 commits are just tidying up the go.mod,go version,tests, and updating dependencies

## Allocators

The first bit of scaffolding is an interface and a plugin for address allocation. Currently this is geared towards ipv6 and prefixes, but should be usable as-is for ipv4 and single addresses (and obviously single ipv6 addresses as well). The interface for now is extremely simple, it has `alloc` and `free`, with `alloc` taking an optional hint of what the client desires.
This allows implementing different strategies to allocate ips, eg sequentially or at random.

The base package for allocators also provides helpers (in `ipcalc.go`) to help with managing ipv6 addresses as 128-bit integers, since `net.IP` itself doesn't have a fixed size and can't be used to index a map. I expect these operations will be basically needed by all the allocators we'll implement, so I put them in the package with the interface declaration

Finally, there is one trivial allocator implemented, which always allocates blocks of the same size. it basically has the logic for an allocator of single addresses, except then you left-shift the addresses to get a prefix instead. Not very interesting, but useful for debugging/prototyping.

## Prefix delegation

The prefix delegation plugin does 2 things with each request (which I think should be mirrored in the other plugins that have to allocate, like range), it tries to reconcile the request with its existing leases, to renew and extend them, and if that fails it allocates new leases and assigns them to the client.

Part of the complexity of the plugin is caused by the fact that it has to implement lease storage (again, which we also do in range) and retrieval, which I think is something we'll have to move in another plugin category like allocators later.

The logic for reconciling requests with existing leases and extending them isn't great, this is something I want to improve asap, but it was starting to be a lot for a single PR. The current one works and mostly does what you expect in "normal" cases.


Natolumin:
#110 reminded me that the allocator is not goroutine-safe, so don't merge this yet :)

insomniacslk:
@Natolumin what about providing a safe lockless string->IP map (or string->interface{}, but I'm not a fan - maybe a common struct) that other plugins can use?

go.mod
```diff
@@ -1,22 +1,22 @@
 module github.com/coredhcp/coredhcp
 
-go 1.12
+go 1.13
```
insomniacslk, line 3: we can probably bump directly to 1.14

plugins/allocators/bitmap/bitmap.go
```diff
@@ -0,0 +1,119 @@
...
+	"net"
+	"strconv"
+
+	"github.com/willf/bitset"
+
+	"github.com/coredhcp/coredhcp/logger"
+	"github.com/coredhcp/coredhcp/plugins/allocators"
+)
+
+var log = logger.GetLogger("plugins/prefix/allocator")
```
insomniacslk, line 25: shouldn't this be `plugins/allocators/bitmap`?

Natolumin:
> what about providing a safe lockless string->IP map (or string->interface{}, but I'm not a fan - maybe a common struct) that other plugins can use?

I think it's definitely out of scope for this PR; I'm interested in the interface and in reuse for IP allocation for now, but to me allocation of addresses (probably not a map most of the time, the problem resembles much more memory allocation so we'll have freelists and buddy allocators etc) and lease storage (essentially a map) are different concerns. We certainly need both, but this PR is already pretty big :)
 
Second, I really don't know what the interface would be for this, string->IP doesn't seem especially useful on its own ? we always have metadata around, but I think it will vary per plugin. I think some kind of interface for storing and retrieving leases, in the same fashion as the Allocator interface, would be very useful and reusable; but the API should probably not be a map, and the implementation may or may not end up being a map (it could go to a database for example)

I think we discussed this before, but I want to make these kinds of things (Allocators, Lease storage, etc) plugins of their own, with different APIs than plugins handling DHCP requests, but that we also register and can switch at will, eg being able to chose putting leases in a db to have multiple servers, or having them in a file so that we can actually restart servers, etc (we have that in `range`, but again it should be common to multiple plugins)

go.mod
Natolumin: Is there something interesting we want from 1.14 ? I bumped for `%w` and error wrapping, I don't know if there are other goodies in 1.14 that we want to use

insomniacslk:
> I think it's definitely out of scope for this PR

I don't know why I asked it on this PR instead of the other issue, sorry about that

> I really don't know what the interface would be for this, string->IP doesn't seem especially useful on its own

Yeah, that's why I said string -> some struct too, that sounds more useful. 

> I want to make these kinds of things (Allocators, Lease storage, etc) plugins of their own

I agree - should we define their interface so that implementations are consistent and transparent?

pmazzini:
I thought you had merge access. I added you now. Feel free to merge.

Natolumin:
I think I wanted to wait for #111 but I guess we can also go ahead and revisit later. 
I'll double-check tomorrow this is in the state I want it to be and merge

Issue #110, opened 2020-07-12
Adding mutex protection in Handler4 Handler6 to avoid concurrent access


gadrenayan:
Hi,

I have been using the coredhcp server for some learning purpose and I was doing some performance tests using 
"**perfdhcp**" tool. I saw there were race conditions hit in the **Handler4** where there were concurrent read write access to the map of lease records. I now know that map is not good for concurrent access. So I tested by adding a minor change of locking this path with a `&sync.mutex`. And I have not hit the race since.

Have you gone through the concurrency consideration in any future releases ?

I saw that the server either simply fails to process any requests when the race hits, or performs smoothly for the entire duration of the tests.
[2020-07-13T01:00:13+05:30]  INFO plugins/range: found IP address 192.168.245.54 for MAC 00:0c:01:02:03:04
concurrent map writes
fatal error: concurrent map writes

root@ngadre-Inspiron-3542:~/go/src/github.com/coredhcp# perfdhcp -l veth11
Running: perfdhcp -l veth11
^C***Rate statistics***
Rate: 0 4-way exchanges/second
***Statistics for: DISCOVER-OFFER***
sent packets: 498055
received packets: 0
drops: 498055

min delay: inf ms
avg delay: Delay summary unavailable! No packets received.

***Statistics for: REQUEST-ACK***
sent packets: 0
received packets: 0
drops: 0

min delay: inf ms
avg delay: Delay summary unavailable! No packets received.

**With Mutex protection:**

root@ngadre-Inspiron-3542:~/vethDHCP# perfdhcp -l veth11
Running: perfdhcp -l veth11
^C***Rate statistics***
Rate: 3360.67 4-way exchanges/second
***Statistics for: DISCOVER-OFFER***
sent packets: 85109
received packets: 85106
drops: 3

min delay: 0.091 ms
avg delay: 1.201 ms
max delay: 34.907 ms
std deviation: 3.124 ms
collected packets: 1

***Statistics for: REQUEST-ACK***
sent packets: 85106
received packets: 85036
drops: 70

min delay: 0.096 ms
avg delay: 1.008 ms
max delay: 28.306 ms
std deviation: 2.407 ms
collected packets: 70

insomniacslk:
Thanks @gadrenayan , this is very useful! We should definitely fix this issue. I'm wondering what would the performance hit of a mutex be, and I would like to explore an alternative approach with goroutine and channels to do this in a lockless manner. I'll run some experiment this week

[...]

PR #112, opened 2020-08-01
Mutex goroutine

ngadre-juniper:
This Pull Request fix for [#110](https://github.com/coredhcp/coredhcp/issues/110#issue-655462377) as well as separating the lease file storage using a goroutine.


Natolumin:
This is mostly superseded by rewrites in other PRs

PR #115, opened 2020-12-25
Add domain search list plugin

wkronmiller:
Hey, 

I went ahead and threw together a quick plugin to use the insomniacslk OptDomainSearch feature to set default DNS search domains. 

I know it works for DHCPv4 but I don't have a DHCPv6 network to test it on. 

I'm happy to make changes, add tests, etc... though I may be a little slow in pushing them out. If someone else wants to take over the PR to flesh it out, that's totally cool with me.

Cheers!


cmds/coredhcp/main.go
```diff
@@ -20,7 +20,6 @@ import (
 
 	"github.com/coredhcp/coredhcp/plugins"
 	pl_dns "github.com/coredhcp/coredhcp/plugins/dns"
-	pl_searchdomains "github.com/coredhcp/coredhcp/plugins/searchdomains"
```
wkronmiller: This change breaks compilation

PR #116, opened 2020-12-25
Add domain search list plugin

wkronmiller:
Add searchdomains plugin to specify a list of default search domains for DHCPv4 and DHCPv6 clients.

### DIFF
--- a/plugins/searchdomains/plugin.go
+++ b/plugins/searchdomains/plugin.go
@@ -34,35 +34,44 @@ var Plugin = plugins.Plugin{
 	Setup4: setup4,
 }
 
-// These are the DHCP options that are added by the plugin.
+// These are the DNS search domains that are set by the plugin.
 // Note that DHCPv4 and DHCPv6 options are totally independent.
 // If you need the same settings for both, you'll need to configure
 // this plugin once for the v4 and once for the v6 server.
-var v4Option dhcpv4.Option
-var v6Option dhcpv6.Option
+var v4SearchList []string
+var v6SearchList []string
+
+// copySlice creates a new copy of a string slice in memory.
+// This helps to ensure that downstream plugins can't corrupt
+// this plugin's configuration
+func copySlice(original []string) []string {
+	copied := make([]string, len(original))
+	copy(copied, original)
+	return copied
+}
 
 func setup6(args ...string) (handler.Handler6, error) {
-	v6Option = dhcpv6.OptDomainSearchList(&rfc1035label.Labels{
-		Labels: args,
-	})
-	log.Println(v6Option)
+	v6SearchList = args
+	log.Printf("Registered domain search list (DHCPv6) %s", v6SearchList)
 	return domainSearchListHandler6, nil
 }
 
 func setup4(args ...string) (handler.Handler4, error) {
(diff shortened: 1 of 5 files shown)


plugins/searchdomains/plugin.go
```diff
@@ -0,0 +1,68 @@
...
+	Setup6: setup6,
+	Setup4: setup4,
+}
+
+// These are the DHCP options that are added by the plugin.
+// Note that DHCPv4 and DHCPv6 options are totally independent.
+// If you need the same settings for both, you'll need to configure
+// this plugin once for the v4 and once for the v6 server.
+var v4Option dhcpv4.Option
+var v6Option dhcpv6.Option
```
Natolumin, line 42: Using global variables for the returned options like this means that if any other plugin modifies the packet in-place, it would cause side-effects for every other packet handled by this plugin.
You'll need to deep-copy the option on every packet in the handler instead, or keep the search list instead of the full option and build the option by `copy`ing the list in the handlers

(for v4 the option is actually serialized to bytes during `UpdateOption` so it'll work, but it's not an API guarantee afaik. For v6 it's not copied at all)

plugins/searchdomains/plugin.go
```diff
@@ -0,0 +1,68 @@
...
+
+func setup6(args ...string) (handler.Handler6, error) {
+	v6Option = dhcpv6.OptDomainSearchList(&rfc1035label.Labels{
+		Labels: args,
+	})
+	log.Println(v6Option)
```
Natolumin, line 48: Could you make this log more explicit ? eg something like "Registered domain search list (DHCPv6): [list of domains]"

plugins/searchdomains/plugin.go
```diff
@@ -0,0 +1,68 @@
...
+	return domainSearchListHandler6, nil
+}
+
+func setup4(args ...string) (handler.Handler4, error) {
+	v4Option = dhcpv4.OptDomainSearch(&rfc1035label.Labels{
+		Labels: args,
+	})
+	log.Println(v4Option)
```
Natolumin, line 56: (same for this one)

Natolumin (commented):
Thanks for the plugin! Couple comments inline. 
In #115 you mentioned you couldn't test on DHCPv6; I can help with that, and we also have some helper scripts using netns in `.travis` that you can use to setup a testbench on linux if you want

plugins/searchdomains/plugin.go
wkronmiller: Done!

plugins/searchdomains/plugin.go
wkronmiller: Done

plugins/searchdomains/plugin.go
wkronmiller: Wow, I didn't consider that. I switched to storing the search list in memory and copying it into a new `Option` each time, per your suggestion.

Natolumin:
Oh gosh I forgot about this; this seems to be working as it is now after testing. If we find other issues later we'll address them then but let's not make you wait for months for a simple plugin like this and merge this

Natolumin (approved):

PR #118, opened 2021-01-27
plugins/prefix: Fix a panic with empty hints

Natolumin:
A missing nilcheck caused prefix allocation to panic for requests without hints

### DIFF
--- a/plugins/prefix/plugin_test.go
+++ b/plugins/prefix/plugin_test.go
@@ -80,3 +80,14 @@ func TestRoundTrip(t *testing.T) {
 			iapd.Options.Prefixes())
 	}
 }
+
+func TestDup(t *testing.T) {
+	_, prefix, err := net.ParseCIDR("2001:db8::/48")
+	if err != nil {
+		panic("bad cidr")
+	}
+	dupPrefix := dup(prefix)
+	if !samePrefix(dupPrefix, prefix) {
+		t.Fatalf("dup doesn't work: got %v expected %v", dupPrefix, prefix)
+	}
+}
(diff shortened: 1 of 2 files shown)


plugins/prefix/plugin_test.go
```diff
@@ -80,3 +80,14 @@ func TestRoundTrip(t *testing.T) {
 			iapd.Options.Prefixes())
 	}
 }
+
+func TestDup(t *testing.T) {
+	_, prefix, err := net.ParseCIDR("2001:db8::/48")
+	if err != nil {
+		panic("bad cidr")
```
insomniacslk, line 87: shouldn't we use `t.Fatalf` or `t.Panicf` here? And more in general we should consolidate everything to use `stretchr/testify`

insomniacslk (approved):
Just one nit, but other than that this LGTM

insomniacslk:
Also not sure what's wrong with TravisCI, I'll migrate to GitHub Actions ASAP

plugins/prefix/plugin_test.go
Natolumin: not really, this isn't the test failing, it's a bug in writing the test (if `net.ParseCIDR("2001:db8::/48")` fails we have issues). I'm using it kind of like rust's `unwrap()`. Maybe there's a better idiom ?

`testify` was only introduced in this repo in the very latest PR. the dhcp lib has been using it for a while but so far barely anything uses it in coredhcp, I'm not opposed to it but I don't think we've actively decided to consolidate on it before this ?

insomniacslk: Some packages use `Must*` versions of functions like this, e.g. `MustParseCIDR`, which basically wrap a `panic` in case of error. Since there's no such function here, and this is a one-off, I'd just leave `panic`. Feel free to merge when you think its' ready :)

PR #119, opened 2021-01-27
Migrate to GitHub Actions

insomniacslk:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: setup environment
+        run: |
+          echo "GOPATH=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+          echo "GO111MODULE=on" >> $GITHUB_ENV
+      - name: build coredhcp
+        run: |
+          set -exu
+          cd $GITHUB_WORKSPACE/src/github.com/${{ github.repository }}/cmds/coredhcp
+          go build
+  coredhcp-generator:
```
Natolumin, line 30: This one is more of a lint than a test I think ? it's just checking that the 2 `main.go` don't drift from each other

.github/workflows/tests.yml
```diff
@@ -0,0 +1,90 @@
...
+          path: src/github.com/${{ github.repository }}
+      - uses: actions/setup-go@v2
+        with:
+          stable: false
+          go-version: ${{ matrix.go }}
+      - name: setup environment
+        run: |
+          echo "GOPATH=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+          echo "GOBIN=$GITHUB_WORKSPACE/bin" >> $GITHUB_ENV
+          echo "GO111MODULE=off" >> $GITHUB_ENV # until we solve this for coredhcp-generator
```
Natolumin, line 48: Any reason to use it like this instead of the `env:` attribute https://docs.github.com/en/actions/reference/environment-variables#about-environment-variables ? (worth adding in a comment if there's a specific reason to do it like this)

.github/workflows/lint.yml
```diff
@@ -0,0 +1,40 @@
...
+    name: checklicenses
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: check license headers
+        run: |
+          set -exu
+          go get -u github.com/u-root/u-root/tools/checklicenses
+          go install github.com/u-root/u-root/tools/checklicenses
+          go run github.com/u-root/u-root/tools/checklicenses -c .ci/checklicenses_config.json
```
Natolumin, line 40: I don't think we need all 3 of these, `go get` already does the install, this should be enough right ?
```suggestion
          go get -u github.com/u-root/u-root/tools/checklicenses
          $(go env GOPATH)/bin/checklicenses -c .ci/checklicenses_config.json
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,60 @@
+name: Build
+
+on: [push, pull_request]
```
Natolumin, line 3: This seems to cause actions to be duplicated; I think it's a known issue (https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012) there are a few workarounds in that thread but they don't seem great

Natolumin (approved):
minor comments but this is nice (and everything is in the github UI instead of having to chase into travis as well

.github/workflows/tests.yml
insomniacslk: I guess it's border-line: we are building coredhcp-generator (which is a test per-se) and running it to build a source file, and then comparing it. I'm ok with moving it under `lint.yml` but it won't be a perfect fit there either

.github/workflows/tests.yml
insomniacslk: Can't remember the details, but one reason is that `env` doesn't let you dynamically construct variables from other variables. I just reused this approach from another configuration that I wrote, which in this case would be only necessary for GOBIN. I can move the other two to `env`, and leave a comment for GOBIN

.github/workflows/lint.yml
insomniacslk: correct, fixing it

.github/workflows/tests.yml
insomniacslk: I am OK with duplicate actions as the least dangerous/ugly thing: otherwise a direct push to `master` without PR could be problematic. A maybe cleaner approach is to have only `master` checked on `push`, and anything else on `pull_request`. What do you think?

.github/workflows/tests.yml
insomniacslk: actually only GO111MODULE=off can be moved to `env`, in which case I'd rather keep all the variables there for consistency

insomniacslk: I'll add a comment

insomniacslk: done

.github/workflows/lint.yml
insomniacslk: done

insomniacslk:
up for review again

.github/workflows/tests.yml
Natolumin: better to have too many than miss one; if you can do the master-on-push everything-else-on-pr that's probably best, but otherwise this is better than missing commits yes

.github/workflows/tests.yml
Natolumin: fair enough let's leave it like this then

Natolumin (approved):

PR #121, opened 2021-02-02
plugins/range: Add some locking and use the allocator interface

Natolumin:
Mixed improvements to overhaul the plugin
 * Use the allocator interface to allocate IPs
 * Focus on correctness by wrapping a lock for now; we'll improve that
   with some more fine-grained concurrency later
 * move things related to file storage of leases to a separate source
   file and add some minimal sanity-check tests
 * Don't open and close the file descriptor on every single packet

### DIFF
--- a/plugins/allocators/bitmap/bitmap_ipv4.go
+++ b/plugins/allocators/bitmap/bitmap_ipv4.go
@@ -19,13 +19,8 @@ import (
 )
 
 var (
-	// ErrNoAddrAvail is returned when we can't allocate an IP
-	// because there's no unallocated space left
-	ErrNoAddrAvail = errors.New("No address available to allocate")
-
-	errNotInRange = errors.New("IP outside of allowed range")
-
-	errInvalidIP = errors.New("Invalid IP passed as input")
+	errNotInRange = errors.New("IPv4 address outside of allowed range")
+	errInvalidIP  = errors.New("invalid IPv4 address passed as input")
 )
 
 // IPv4Allocator allocates IPv4 addresses, tracking utilization with a bitmap
@@ -33,16 +28,18 @@ type IPv4Allocator struct {
 	start uint32
 	end   uint32
 
+	// This bitset implementation isn't goroutine-safe, we protect it with a mutex for now
+	// until we can swap for another concurrent implementation
 	bitmap *bitset.BitSet
 	l      sync.Mutex
 }
 
 func (a *IPv4Allocator) toIP(offset uint32) net.IP {
 	if offset > a.end-a.start {
-		return nil
+		panic("BUG: offset out of bounds")
 	}
 
-	r := make(net.IP, 4)
+	r := make(net.IP, net.IPv4len)
 	binary.BigEndian.PutUint32(r, a.start+offset)
 	return r
 }
@@ -78,7 +75,7 @@ func (a *IPv4Allocator) Allocate(hint net.IPNet) (n net.IPNet, err error) {
 		// Then any available address
 		avail, ok := a.bitmap.NextClear(0)
 		if !ok {
-			return n, ErrNoAddrAvail
+			return n, allocators.ErrNoAddrAvail
 		}
 		next = avail
 	}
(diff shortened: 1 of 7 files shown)


plugins/allocators/bitmap/bitmap_ipv4.go
```diff
@@ -0,0 +1,125 @@
...
+	errInvalidIP = errors.New("Invalid IP passed as input")
+)
+
+// IPv4Allocator allocates IPv4 addresses, tracking utilization with a bitmap
+type IPv4Allocator struct {
+	start uint32
+	end   uint32
+
+	bitmap *bitset.BitSet
+	l      sync.Mutex
```
Natolumin, line 37: I'll move this to use a concurrent bitset at some point in the future, https://github.com/tmthrgd/go-bitset seemed nice for example, so that we can remove the mutex here (and in the v6 allocator as well)

plugins/allocators/bitmap/bitmap_ipv4.go
```diff
@@ -0,0 +1,125 @@
...
+
+	intIP := binary.BigEndian.Uint32(ip.To4())
+	if intIP < a.start || intIP > a.end {
+		return 0, errNotInRange
+	}
+
+	return uint(intIP - a.start), nil
+}
+
+// Allocate reserves an IP for a client
```
insomniacslk, line 63: can you add a comment on how the hint works for the user?

plugins/allocators/bitmap/bitmap_ipv4.go
```diff
@@ -0,0 +1,125 @@
...
+func (a *IPv4Allocator) Allocate(hint net.IPNet) (n net.IPNet, err error) {
+	n.Mask = net.CIDRMask(32, 32)
```
insomniacslk, line 65: if we always overwrite the mask, why not using `net.IP` instead of `net.IPNet` in the interface?

plugins/allocators/bitmap/bitmap_ipv4.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, fmt.Errorf("Invalid IPv4s given to create the allocator: [%s,%s]", start, end)
+	}
+
+	alloc := IPv4Allocator{
+		start: binary.BigEndian.Uint32(start.To4()),
+		end:   binary.BigEndian.Uint32(end.To4()),
+	}
+
+	if alloc.start > alloc.end {
+		return nil, errors.New("No IPs in the given range to allocate")
```
insomniacslk, line 120: nit: error strings should should not be capitalized

insomniacslk (commented):
Should we also move the allocators from `plugins` to `pkg`? We should keep only actual plugins there

plugins/allocators/bitmap/bitmap_ipv4.go
Natolumin: This is just the implementation, there's a user-level explanation in the [interface](https://github.com/coredhcp/coredhcp/blob/f4b74e59cadf52f325dc856d7b3fa6295bb112cc/plugins/allocators/allocator.go#L21)

plugins/allocators/bitmap/bitmap_ipv4.go
Natolumin: The interface is shared between v4 and v6, and v6 uses masks for prefix delegation

plugins/allocators/bitmap/bitmap_ipv4.go
```diff
@@ -0,0 +1,125 @@
...
+	start uint32
+	end   uint32
+
+	bitmap *bitset.BitSet
+	l      sync.Mutex
+}
+
+func (a *IPv4Allocator) toIP(offset uint32) net.IP {
+	if offset > a.end-a.start {
+		return nil
```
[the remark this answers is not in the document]
Natolumin: while toOffset takes user-supplied data, toIP is the opposite and only handles output data from the plugin; if anything this should log an error or even panic, this is always a bug

plugins/allocators/bitmap/bitmap_ipv4.go
Natolumin: This specific allocator has 2 implementations for each protocol version but others may not need to

plugins/allocators/bitmap/bitmap_ipv4.go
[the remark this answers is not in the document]
insomniacslk: sounds good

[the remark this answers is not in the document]
insomniacslk: at this point, shouldn't it `panic` instead of returning nil?

plugins/allocators/bitmap/bitmap_ipv4.go
insomniacslk: right

insomniacslk (approved):
LGTM

PR #122, opened 2021-02-03
ci: Map paths in codecov to paths in the repo

Natolumin:
Codecov uses the location of files on disk during coverage, which in our action has an additional prefix compared to the path in the repo, making some of the stats in codecov not work.
https://docs.codecov.io/docs/fixing-paths explains how to address that

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,4 @@
+# Map file paths during tests to the canonical path in the repo
+fixes:
+    # Removes the prefix we use during github actions
+    - "src/github.com/coredhcp/coredhcp/::"


Natolumin:
it works! Compare: 
master - 404: https://codecov.io/gh/coredhcp/coredhcp/src/master/src/github.com/coredhcp/coredhcp/plugins/searchdomains/plugin.go
this branch - annotated source: https://codecov.io/gh/coredhcp/coredhcp/src/fc69388782683b0389e4b0eaa861f10cd8d6d093/plugins/searchdomains/plugin.go

insomniacslk (approved):
Awesome, thanks!

PR #123, opened 2021-02-12
Send unicast ACK

dza1:
Add function that sends an ACK with a Broadcast IP, but a destination MAC address

### DIFF
--- a/server/handle.go
+++ b/server/handle.go
@@ -172,7 +172,7 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
 			}
 		}
 
-		if useEthernet == true {
+		if useEthernet {
 			intf, err := net.InterfaceByIndex(woob.IfIndex)
 			if err != nil {
 				log.Printf("MainHandler4: Can not get Interface for index %d %v", woob.IfIndex, err)
--- a/server/handle.go
+++ b/server/handle.go
@@ -139,6 +139,7 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
 		}
 	}
 
+	useEthernet := false
 	if resp != nil {
 		var peer *net.UDPAddr
 		if !req.GatewayIPAddr.IsUnspecified() {
@@ -152,9 +153,8 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
 			peer = &net.UDPAddr{IP: net.IPv4bcast, Port: dhcpv4.ClientPort}
 		} else {
 			//sends a layer2 frame so that we can define the destination MAC address
-			sendEthernet(l.Interface, resp)
-			return
-
+			peer = &net.UDPAddr{IP: net.IPv4bcast, Port: dhcpv4.ClientPort}
+			useEthernet = true
 		}
 
 		var woob *ipv4.ControlMessage
@@ -171,8 +171,17 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
 				log.Errorf("HandleMsg4: Did not receive interface information")
 			}
 		}
-		if _, err := l.WriteTo(resp.ToBytes(), woob, peer); err != nil {
-			log.Printf("MainHandler4: conn.Write to %v failed: %v", peer, err)
+
+		if useEthernet == true {
+			intf, err := net.InterfaceByIndex(woob.IfIndex)
+			if err != nil {
+				log.Printf("MainHandler4: Can not get Interface for index %d %v", woob.IfIndex, err)
+			}
+			sendEthernet(*intf, resp)
+		} else {
+			if _, err := l.WriteTo(resp.ToBytes(), woob, peer); err != nil {
+				log.Printf("MainHandler4: conn.Write to %v failed: %v", peer, err)
+			}
 		}
 	} else {
 		log.Print("MainHandler4: dropping request because response is nil")
--- a/server/handle.go
+++ b/server/handle.go
@@ -139,8 +139,8 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
 		}
 	}
 
-	useEthernet := false
 	if resp != nil {
+		useEthernet := false
 		var peer *net.UDPAddr
 		if !req.GatewayIPAddr.IsUnspecified() {
 			// TODO: make RFC8357 compliant
@@ -153,13 +153,13 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
 			peer = &net.UDPAddr{IP: net.IPv4bcast, Port: dhcpv4.ClientPort}
 		} else {
 			//sends a layer2 frame so that we can define the destination MAC address
-			peer = &net.UDPAddr{IP: net.IPv4bcast, Port: dhcpv4.ClientPort}
+			peer = &net.UDPAddr{IP: resp.YourIPAddr, Port: dhcpv4.ClientPort}
 			useEthernet = true
 		}
 
 		var woob *ipv4.ControlMessage
-		if peer.IP.Equal(net.IPv4bcast) || peer.IP.IsLinkLocalUnicast() {
-			// Direct broadcasts and link-local to the interface the request was
+		if peer.IP.Equal(net.IPv4bcast) || peer.IP.IsLinkLocalUnicast() || useEthernet {
+			// Direct broadcasts, link-local and layer2 unicasts to the interface the request was
 			// received on. Other packets should use the normal routing table in
 			// case of asymetric routing
 			switch {
@@ -176,6 +176,7 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
 			intf, err := net.InterfaceByIndex(woob.IfIndex)
 			if err != nil {
 				log.Errorf("MainHandler4: Can not get Interface for index %d %v", woob.IfIndex, err)
+				return
 			}
 			err = sendEthernet(*intf, resp)
 			if err != nil {
--- a/server/handle.go
+++ b/server/handle.go
@@ -144,13 +144,14 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
 		if !req.GatewayIPAddr.IsUnspecified() {
 			// TODO: make RFC8357 compliant
 			peer = &net.UDPAddr{IP: req.GatewayIPAddr, Port: dhcpv4.ServerPort}
-		} else if resp.MessageType() == dhcpv4.MessageTypeNak || req.MessageType() == dhcpv4.MessageTypeDiscover {
+		} else if resp.MessageType() == dhcpv4.MessageTypeNak {
 			peer = &net.UDPAddr{IP: net.IPv4bcast, Port: dhcpv4.ClientPort}
 		} else if !req.ClientIPAddr.IsUnspecified() {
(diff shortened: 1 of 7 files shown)


server/sendEthernet.go
```diff
@@ -0,0 +1,93 @@
...
+			Data:   data,
+			Length: uint8(len(data)),
+		})
+		return
+	}
+
+	//Add all option to the layer struct
+	for key, element := range resp.Options {
+		appendOption(layers.DHCPOpt(key), []byte(element))
+	}
```
Natolumin, line 66: Since the dhcp library we use already has serialization code for dhcp, can we replace this with just the payload as a byte slice (from `resp.ToBytes()`) ? or does this absolutely need to use the layers from gopacket for serialization ?

One reason I'd like to avoid this is that if we end up with 2 different serialization codepaths it'd make debugging painful

server/sendEthernet.go
```diff
@@ -0,0 +1,93 @@
...
+
+//function from https://gist.github.com/corny/5e4e3f8e6f2395726e46c3db9db17f12#file-dhcp_discover-go
+import (
+	"encoding/binary"
+	"net"
+	"time"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+	"github.com/google/gopacket/pcap"
```
Natolumin, line 11: For reference, how much of these are unix (or linux) only ? I believe we already have a couple parts where the code is linux-only but it'd be nice to document it as much as possible for if/when we try to broaden supported platforms

server/handle.go
```diff
@@ -151,14 +151,10 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
...
 			peer = &net.UDPAddr{IP: net.IPv4bcast, Port: dhcpv4.ClientPort}
 		} else {
-			// FIXME: we're supposed to unicast to a specific *L2* address, and an L3
-			// address that's not yet assigned.
-			// I don't know how to do that with this API...
-			//peer = &net.UDPAddr{IP: resp.YourIPAddr, Port: dhcpv4.ClientPort}
-			log.Warn("Cannot handle non-broadcast-capable unspecified peers in an RFC-compliant way. " +
-				"Response will be broadcast")
+			//sends a layer2 frame so that we can define the destination MAC address
+			sendEthernet(l.Interface, resp)
```
Natolumin, line 155: `l.Interface` here isn't necessarily non-nil if the configuration doesn't restrict coredhcp to listen on a specific interface; you'll need to use the same logic as lines 161-174 to find the right interface for the return packet

server/sendEthernet.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	err := gopacket.SerializeLayers(buf, opts, &eth, &ip, &udp, &dhcp)
+	if err != nil {
+		log.Errorf("Can not serialize layer for AKN: %v, err")
+	}
+	data := buf.Bytes()
+
+	log.Debugf("sending %d bytes for AKN", len(data))
+
+	handle, err := pcap.OpenLive(iface.Name, 1024, false, time.Second)
```
Natolumin, line 82: Can the pcap library open interfaces by index ? the name of the interface could change between when we receive the packet and when we send the response (that's very unlikely so it's not very important, but if we can avoid it at all it's better)

Natolumin (commented):
Wow thanks, this is pretty great!

I have a few comments inline but I'd be happy to merge something like this
Also maybe a stupid question: what are you using "AKN" as an abbreviation for ?

server/sendEthernet.go
dza1: Good Point! 
Now I use resp.ToBytes()

server/handle.go
dza1: Should be OK now.

dza1:
Thanks for the quick review. I used a wrong abbreviation for ACK, so I changed it now. But it is also used for DHCPOFFE, so it is probably a bad title. I made now a few changing according to your suggestions, so could you please take a look at it again.

server/sendEthernet.go
dza1: According https://pkg.go.dev/github.com/google/gopacket@v1.1.19/pcap, you need to install winpcap or npcap to use it on Windows, but I didn't try.

server/sendEthernet.go
dza1: I didn't find a possibility to open the interface by index, but now the Interface name is reed out before the `sendEthernet()` call. So the time between reading the name and use it is really short.

Natolumin:
There's a last question to be addressed here; I'm not asking you to do so, it's just something we have to think about and decide before going forward with this. 
Some of our usecases/users probably want to avoid CGO if at all possible; and even with cgo, requiring libpcap to be available makes this harder to build into a static easy-to-deploy binary.

A cop-out or at least mitigation would be to have a build tag to enable or disable this binding, falling back to the previous, or some other behaviour

Nothing to do immediately, but we'll need to have a bit of discussion/time to figure out what we want to do

dza1:
To avoid CGO and libpcap, we can use a Socket, which one we can open with the `syscall` library. I tried it already and it is working. If you prefer that, I can commit it or open another pull request.

Natolumin:
That would be great yes, you can add it to this PR

dza1:
In last commit we use now a socket for the layer2 communication, so there shouldn't be any CGO involved. I also added an error return value for the `sendEthernet()` function.

server/handle.go
```diff
@@ -139,6 +139,7 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
 		}
 	}
 
+	useEthernet := false
```
Natolumin, line 142: nit: move this into the `if{` block

server/handle.go
```diff
@@ -151,14 +152,9 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
...
 		} else {
-			// FIXME: we're supposed to unicast to a specific *L2* address, and an L3
-			// address that's not yet assigned.
-			// I don't know how to do that with this API...
-			//peer = &net.UDPAddr{IP: resp.YourIPAddr, Port: dhcpv4.ClientPort}
-			log.Warn("Cannot handle non-broadcast-capable unspecified peers in an RFC-compliant way. " +
-				"Response will be broadcast")
-
+			//sends a layer2 frame so that we can define the destination MAC address
 			peer = &net.UDPAddr{IP: net.IPv4bcast, Port: dhcpv4.ClientPort}
```
Natolumin, line 156: https://tools.ietf.org/html/rfc2131#section-4.1(last paragraph) In this case this should be sent to the yiaddr, not broadcast
```suggestion
			peer = &net.UDPAddr{IP: resp.YourIPAddr, Port: dhcpv4.ClientPort}
```

server/handle.go
```diff
@@ -175,10 +171,21 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
...
 			}
 		}
-		if _, err := l.WriteTo(resp.ToBytes(), woob, peer); err != nil {
-			log.Printf("MainHandler4: conn.Write to %v failed: %v", peer, err)
-		}
 
+		if useEthernet {
+			intf, err := net.InterfaceByIndex(woob.IfIndex)
+			if err != nil {
+				log.Errorf("MainHandler4: Can not get Interface for index %d %v", woob.IfIndex, err)
```
Natolumin, line 178: Need to return from the function otherwise there will be a nil pointer dereference on l180
```suggestion
				log.Errorf("MainHandler4: Can not get Interface for index %d %v", woob.IfIndex, err)
				return
```

server/sendEthernet.go
```diff
@@ -0,0 +1,94 @@
+// Copyright 2018-present the CoreDHCP Authors. All rights reserved
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
```
Natolumin, line 4: Even though that' probably the only thing we support today, we should explicitly not build it otherwise
```suggestion

// +build linux
```

server/sendEthernet.go
```diff
@@ -0,0 +1,94 @@
...
+	eth := layers.Ethernet{
+		EthernetType: layers.EthernetTypeIPv4,
+		SrcMAC:       iface.HardwareAddr,
+		DstMAC:       resp.ClientHWAddr,
+	}
+	ip := layers.IPv4{
+		Version:  4,
+		TTL:      64,
+		SrcIP:    resp.ServerIPAddr,
+		DstIP:    net.IPv4bcast,
```
Natolumin, line 32:
```suggestion
		DstIP:    resp.YourIPAddress,
```

server/handle.go
Natolumin: However that makes the logic under it to select the interface index incorrect, so I'd say lift the logic to choose an interface index out into its own function here

Natolumin (commented):
Sorry for the long wait! There's only a few small things to take care of now, but I can add them myself if you don't have time to work on this, and then I'll merge it
I'll also check the build failure because I'm not quite sure what's going on there

server/sendEthernet.go
dza1: I added your suggestions now. 
About the build failure, I don't know what the problem is. Would be great if you can take a look.

dza1:
I added your suggestions now.
About the build failure, I don't know what the problem is. Would be great if you can take a look.

Natolumin (approved):
LGTM with the last changes. I'll run some tests and merge this soon unless there are any other objections. The build failure is fixed with #126

server/sendEthernet.go
```diff
@@ -0,0 +1,96 @@
...
+
+	eth := layers.Ethernet{
+		EthernetType: layers.EthernetTypeIPv4,
+		SrcMAC:       iface.HardwareAddr,
+		DstMAC:       resp.ClientHWAddr,
+	}
+	ip := layers.IPv4{
+		Version:  4,
+		TTL:      64,
+		SrcIP:    resp.ServerIPAddr,
```
Natolumin, line 33: One maybe surprising thing is that is server_id is not set, this ends up sending a packet with source address 0.0.0.0 which is probably surprising

elad:
@Natolumin could you please merge this PR? Since DHCP assigns IP addresses, this change allows responses to be sent over Ethernet, which is required in scenarios where that's the only way to communicate back to the client. 🙂 I confirm that it works well.

Note that for this to build cleanly on macOS, one approach is to add a file (`sendEthernet_darwin.go`) with the following:

```go
// +build darwin

package server

import (
	"fmt"
	"net"

	"github.com/insomniacslk/dhcp/dhcpv4"
)

func sendEthernet(iface net.Interface, resp *dhcpv4.DHCPv4) error {
	return fmt.Errorf("not implemented")
}
```

Natolumin:
Ah yes sorry, let's merge it, I forgot

Can you expand a bit on this ?
> where that's the only way to communicate back to the client

What's your setup where you can't broadcast to a client but you're not using a relay ? This might be something we want to keep in mind if this is a common configuration (but I'm not sure how you'd manage to have that issue, hence the question)

elad:
> What's your setup where you can't broadcast to a client but you're not using a relay ?

My setup is a virtual network where we prefer to use unicast and may not have broadcast.

Natolumin:
> My setup is a virtual network where we prefer to use unicast and may not have broadcast.

The virtual network is still ethernet right, since this PR only supports ethernet anyway ?
I still don't understand how you can avoid broadcast, it's mandatory both for ARP and DHCPDISCOVER requests (unless you use a relay but in that case this whole PR is irrelevant to your situation)

I understand preferring unicast of course (and so does the RFC), but I'm failing to see how you get to a point where you **can't** use broadcast. If I could understand your setup I could figure out whether we need a way to completely disable the broadcast fallback as well

elad:
Our virtual network implements only the functionality we need, including specific flows to make ARP and DHCPDISCOVER work without relying on broadcast or relays.

PR #124, opened 2021-02-14
Inject ARP table entry to respond with unicast

DSpeichert:
This change allows coredhcp to respond using unicast to a client, who's not known on the network yet (in ARP table).

The method here is closely based on the corresponding behavior of dnsmasq, that can be found [here](http://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=blob;f=src/dhcp.c;h=97324f2377acc27f8997215523cb9c910fa92f7f;hb=cfcafdd27c74dc187fe96a9cfa88b1aef53540a0#l418). Server will simply add the MAC/IP entry to the ARP table manually, prior to sending DHCP response packet.

It's obviously relying on unix being the platform but I believe other parts of coredhcp already limit it to run on unix-like only.

It is an alternative approach to what has been proposed in #114, and that looks like a more native/pure Go solution. Since I had that written already, I'm still offering this to gather feedback. I'm looking forward to see what makes more sense for coredhcp.

### DIFF
--- a/server/arp.go
+++ b/server/arp.go
@@ -1,0 +1,70 @@
+package server
+
+import (
+	"golang.org/x/sys/unix"
+	"net"
+	"os"
+	"syscall"
+	"unsafe"
+)
+
+// ARP Flag values
+// these are not in golang.org/x/sys/unix
+const (
+	// completed entry (ha valid)
+	ATF_COM = 0x02
+	// permanent entry
+	ATF_PERM = 0x04
+	// publish entry
+	ATF_PUBL = 0x08
+	// has requested trailers
+	ATF_USETRAILERS = 0x10
+	// want to use a netmask (only for proxy entries)
+	ATF_NETMASK = 0x20
+	// don't answer this addresses
+	ATF_DONTPUB = 0x40
+)
+
+// https://man7.org/linux/man-pages/man7/arp.7.html
+type arpReq struct {
+	ArpPa   syscall.RawSockaddrInet4
+	ArpHa   syscall.RawSockaddr
+	Flags   int32
+	Netmask syscall.RawSockaddr
+	Dev     [16]byte
+}
+
+func InjectArp(ip net.IP, mac net.HardwareAddr, flags int32, dev string) (err error) {
+	fd, err := unix.Socket(unix.AF_INET, unix.SOCK_DGRAM, unix.IPPROTO_UDP)
+	if err != nil {
+		return
+	}
+	f := os.NewFile(uintptr(fd), "")
+	defer f.Close()
+
+	return InjectArpFd(uintptr(fd), ip, mac, flags, dev)
+}
+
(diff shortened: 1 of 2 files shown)


Natolumin:
Thanks a lot for this alternate proposition; 

There's a couple differences from #123 that make the choice not obvious at all here: 
 * This requires `CAP_NET_ADMIN` on linux I believe, while sending the raw packet requires `CAP_NET_RAW` (a "smaller" capability)
 *  This modifies the host state/network configuration (what's the TTL on that ARP table entry we're adding btw ?) which I'm not very comfortable having as a default behavior / can make things harder to debug
 * This requires having a route to the destination IP as well, which is *probably* not an issue but I'm sure I could come up with a pathological network configuration that doesn't have that

otoh, this doesn't require any cgo - but if we find a way to send raw packets without requiring cgo (can't be that hard right ?) then I still think sending raw packets is nicer because of those points.

Of course, no decision's been made here yet, those are just my opinions on the question and we need to think about this a bit more

DSpeichert:
You are right about permissions - it does require CAP_NET_ADMIN. 

I don't believe there is a TTL on that entry, based on my research, there is an option to make a permanent entry - or not. So I believe that this entry lives and behaves just as if it was learned from the wire.

It does require knowing a device name - this syscall won't work without it. I also saw your comment about the possibility of device name change (while unlikely) and using the interface index. That also doesn't seem possible with this syscall.

In general, I've studied the actual implementation of it [here](https://github.com/torvalds/linux/blob/8cf8821e15cd553339a5b48ee555a0439c2b2742/net/ipv4/arp.c#L1179) and [here](https://github.com/torvalds/linux/blob/8cf8821e15cd553339a5b48ee555a0439c2b2742/net/ipv4/arp.c#L1024) to figure out how it seems to work. Mapping this syscall to go wasn't trivial, as I find some of the struct fields' names misleading. Some [prior art](https://www.unix.com/302447674-post3.html) helped a lot too.

In terms of the route, I think it requires having an on-network route (it's still L2 if we care about ARP and MAC), which is the same requirement as for broadcast, so the requirements are basically the same.

I generally would prefer sending raw packets instead and I like #123 for that a lot. I've briefly investigated that approach and realized that it would take more code and more libraries to do. Therefore, for my use case, I've settled for simply copying what existing DHCP servers do (hence my reference to dnsmasq specifically), which results in fairly concise code. No cgo indeed. It also easily falls back to previous behavior.

Let me know if you want me to look at the linter comments or take `arp.go` out into a separate package, etc. For now, this is simply in a shape that works.

Natolumin:
Hey, thanks again for this and sorry for taking so long to look at it.
In the end, since we've made #123 not require cgo, even with a bit more libraries required we'll go with it instead. Additionally I think when we decide to take a proper look at portability to non-linux it'll also be a bit easier to port sending a raw packet than interfacing with the arp table for each OS we want, so I'll close this

Please don't hesitate if you have other ideas for coredhcp though!

PR #125, opened 2021-03-10
allocators/bitmap: Fix address usage check in ipv4 allocator

toshic:


plugins/allocators/bitmap/bitmap_ipv4.go
```diff
@@ -69,7 +69,7 @@ func (a *IPv4Allocator) Allocate(hint net.IPNet) (n net.IPNet, err error) {
 
 	var next uint
 	// First try the exact match
-	if a.bitmap.Test(hintOffset) {
+	if !a.bitmap.Test(hintOffset) {
```
Natolumin, line 72: and it'll always test this as well if the hint is invalid. I'll send a follow-up patch to test the hint only if it's valid

Natolumin (approved):
:facepalm: that's a pretty stupid bug, thank you for the fix

I'll have a look at the -generator build failure, I think it's also broken on master, and merge this

Natolumin:
The build failure is definitely unrelated and will be taken care of in #126, I'll merge this now

toshic:
Thank you so much for such a fast merge!

PR #126, opened 2021-03-10
ci: Build coredhcp-generator output with modules

Natolumin:
Viper doesn't support running from gopath at all anymore
(https://github.com/spf13/viper/blob/master/TROUBLESHOOTING.md#cannot-find-package)
causing the hcl import error we've seen in PRs.
We can actually just init a stub go module here, and let whichever of
our dependencies specify their own needs in their own go.mod files. This
ends up resolving to "latest" for all our dependencies, which is roughly
equivalent to not having modules at all

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,16 +48,15 @@ jobs:
           # trick.
           echo "GOPATH=$GITHUB_WORKSPACE" >> $GITHUB_ENV
           echo "GOBIN=$GITHUB_WORKSPACE/bin" >> $GITHUB_ENV
-          echo "GO111MODULE=off" >> $GITHUB_ENV # until we solve this for coredhcp-generator
       - name: build coredhcp-generator
         run: |
           set -exu
           cd "${GITHUB_WORKSPACE}"/src/github.com/${{ github.repository }}/cmds/coredhcp-generator
-          GO111MODULE=on go build
+          go build
           builddir=$(./coredhcp-generator -f core-plugins.txt)
           cd "${builddir}"
           ls -l
-          go get -v ./...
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -48,16 +48,15 @@ jobs:
...
         run: |
           set -exu
           cd "${GITHUB_WORKSPACE}"/src/github.com/${{ github.repository }}/cmds/coredhcp-generator
-          GO111MODULE=on go build
+          go build
           builddir=$(./coredhcp-generator -f core-plugins.txt)
           cd "${builddir}"
           ls -l
-          go get -v ./...
+          go mod init "coredhcp"
```
insomniacslk, line 59: this is OK for the CI - otherwise the correct thing would be "go mod init github.com/coredhcp/coredhcp". I wonder if doing otherwise might break anything in the future when we will move to modules

insomniacslk (approved):

.github/workflows/build.yml
Natolumin: actually `go mod init github.com/coredhcp/coredhcp` breaks it, because it then expects all the imports from `github.com/coredhcp/coredhcp` to be in the current directory. But this directory only has the main entrypoint which is normally in `github.com/coredhcp/coredhcp/cmds/coredhcp`, so it doesn't manage to import anything

to be able to use `go mod init github.com/coredhcp/coredhcp` we'd need to build the generated `coredhcp.go` in-place inside the source directory, which is more painful
I could maybe rename the module as "local-coredhcp-entrypoint" or something more explicit like that ?

insomniacslk: ah got it, thanks for the explanation!

PR #127, opened 2021-03-17
Created sleep plugin

insomniacslk:
The `sleep` plugin introduces a delay in the chain of response to a DHCP
transaction.

### DIFF
--- a/cmds/coredhcp/main.go
+++ b/cmds/coredhcp/main.go
@@ -27,6 +27,7 @@ import (
 	pl_router "github.com/coredhcp/coredhcp/plugins/router"
 	pl_searchdomains "github.com/coredhcp/coredhcp/plugins/searchdomains"
 	pl_serverid "github.com/coredhcp/coredhcp/plugins/serverid"
+	pl_sleep "github.com/coredhcp/coredhcp/plugins/sleep"
 
 	"github.com/sirupsen/logrus"
 	flag "github.com/spf13/pflag"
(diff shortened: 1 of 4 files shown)


insomniacslk:
also added a `go mod edit` call to replace the go.mod references to point to the local source code rather than downloading them. The error was
```
coredhcp.go:30:2: module github.com/coredhcp/coredhcp@latest found (v0.0.0-20210316185101-4f97fe7ae569), but does not contain package github.com/coredhcp/coredhcp/plugins/sleep
```
which, if I interpret correctly, means that after #126 we were building coredhcp-generator by fetching the code from master instead of using the local, patched sources

cmds/coredhcp/main.go
```diff
@@ -27,6 +27,7 @@ import (
 	pl_router "github.com/coredhcp/coredhcp/plugins/router"
 	pl_searchdomains "github.com/coredhcp/coredhcp/plugins/searchdomains"
 	pl_serverid "github.com/coredhcp/coredhcp/plugins/serverid"
+	pl_sleep "github.com/coredhcp/coredhcp/plugins/sleep"
```
Natolumin, line 30: It's probably fine since it's very small and has no dependencies or anything, but at the same time it's such a niche usecase, is there any point having it included by default ?

Maybe we should think about a checklist for choosing which plugins to build by default

Natolumin (approved):

Natolumin:
> means that after #126 we were building coredhcp-generator by fetching the code from master instead of using the local, patched sources

:facepalm: yes indeed; thanks for catching it :)

cmds/coredhcp/main.go
insomniacslk: I included it because it's really small, but I agree with you that we need a clear boundary on what's admissible and what not in the "included batteries". I will merge this for now, we can change our mind later anyway

PR #129, opened 2021-04-24
server: Reuse the raw socket

Natolumin:
Creating, then destroying a raw socket for every packet is very heavy. We can open the socket and keep it when first opening the (udp) listening socket, so that we are able to reuse the raw socket for further packets

Instead of managing the packet socket with manual syscalls, we can use the https://github.com/mdlayher/raw library (which we already imported before anyway) which will give us portability for free.
This does however mean that we cannot support dynamically setting the interface differently for each packet (at the moment) as the library requires an interface when binding to the socket; so this requires binding to a specific interface at startup (rather than relying on out-of-band data when receiving packets to choose the outgoing interface).
For DHCP servers this is a very common behavior (and we even required it until not that long ago) so I'm happy to take that tradeoff until we can update the upstream library to support the other usecase if needed.

Finally, if we can't setup the raw socket, we can then fallback to sending over broadcast (which is allowed by the RFC), ensuring that the server can keep working without for example CAP_NET_RAW


insomniacslk:
> For DHCP servers this is a very common behavior (and we even required it until not that long ago) so I'm happy to take that tradeoff until we can update the upstream library to support the other usecase if needed.

I am OK with this behaviour

PR #131, opened 2021-05-12
Add stub to fix build on macOS

elad:


Natolumin:
Afaict (although I can't test) #129 should also fix the build, without causing a runtime error; can you test that branch ?

elad:
It doesn’t seem like it:

```
$ git show --summary
commit 755c9769cbd1291574f8397f7ef1a2ac5ee9134c (HEAD -> raw_socket, origin/raw_socket)
Author: Anatole Denis <anatole@unverle.fr>
Date:   Sat Apr 24 15:57:16 2021 +0200

    server: Select an appropriate IP address for L2 unicast
    
    When deferring to the system routing, the ip address is automatically
    selected by the kernel.
    We were previously using the message's ServerIPAddr, but that is
    incorrect as a source address since that is the address of the next
    server, and not necessarily the address of the current server or an
    address that is routable by the client
    
    Signed-off-by: Anatole Denis <anatole@unverle.fr>

$ go build ./...
# github.com/coredhcp/coredhcp/server
server/handle.go:182:11: l.sendEthernet undefined (type *listener4 has no field or method sendEthernet)
server/serve.go:152:12: l4.tryOpenRawSock undefined (type *listener4 has no field or method tryOpenRawSock)
$ 
```

hardikdr:
Is there a chance this will go in any soon?

damyan:
@elad Could you, kindly, rebase your branch on latest coredhcp main commit, so we can use your fix temporarily?

PR #132, opened 2021-08-17
added staticroute plugin

skoef:
for configuring one or more classless static routes with DHCP option 121

### DIFF
--- a/cmds/coredhcp/config.yml.example
+++ b/cmds/coredhcp/config.yml.example
@@ -156,3 +156,10 @@ server4:
         # * lease duration can be given in any format understood by go's
         # "ParseDuration": https://golang.org/pkg/time/#ParseDuration
         - range: leases.txt 10.10.10.100 10.10.10.200 60s
+
+        # staticroute advertises additional routes the client should install in
+        # its routing table as described in RFC3442
+        # - staticroute: <destination>,<gateway> [<destination>,<gateway> ...]
+        # where destination should be in CIDR notation and gateway should be
+        # the IP address of the router through which the destination is reachable
+        # - staticroute: 10.20.20.0/24,10.10.10.1
--- a/cmds/coredhcp-generator/core-plugins.txt
+++ b/cmds/coredhcp-generator/core-plugins.txt
@@ -9,3 +9,4 @@ github.com/coredhcp/coredhcp/plugins/router
 github.com/coredhcp/coredhcp/plugins/serverid
 github.com/coredhcp/coredhcp/plugins/searchdomains
 github.com/coredhcp/coredhcp/plugins/sleep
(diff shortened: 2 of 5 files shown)


skoef:
@Natolumin you're absolutely right, didn't look too close to that library. I've rewritten the plugin using that dhcpv4.Route instead!

cmds/coredhcp/config.yml.example
```diff
@@ -156,3 +156,9 @@ server4:
         # * lease duration can be given in any format understood by go's
         # "ParseDuration": https://golang.org/pkg/time/#ParseDuration
         - range: leases.txt 10.10.10.100 10.10.10.200 60s
+
+        # optional static routes could be defined as well
+        # - staticroute: <destination>,<gateway> <destination>,<gateway>
+        # where destination should be in CIDR notation and gateway should be
+        # the IP address of the router through which the destination is reachable
```
Natolumin, line 163: Maybe worth adding a mention here that local subnet routes can be specified with a gateway of `0.0.0.0` (from https://datatracker.ietf.org/doc/html/rfc3442#page-4 "Local Subnet Routes")

cmds/coredhcp/config.yml.example
```diff
@@ -156,3 +156,9 @@ server4:
         # * lease duration can be given in any format understood by go's
         # "ParseDuration": https://golang.org/pkg/time/#ParseDuration
         - range: leases.txt 10.10.10.100 10.10.10.200 60s
+
+        # optional static routes could be defined as well
+        # - staticroute: <destination>,<gateway> <destination>,<gateway>
```
Natolumin, line 161: nit: make it slightly more obvious that this is one or more entries:
```suggestion
        # - staticroute: <destination>,<gateway> [<destination>,<gateway> ...]
```

cmds/coredhcp/config.yml.example
```diff
@@ -156,3 +156,9 @@ server4:
         # * lease duration can be given in any format understood by go's
         # "ParseDuration": https://golang.org/pkg/time/#ParseDuration
         - range: leases.txt 10.10.10.100 10.10.10.200 60s
+
+        # optional static routes could be defined as well
```
Natolumin, line 160: nit: this should follow the same style as other plugin descriptions:
```suggestion
        # staticroute advertises additional routes the client should install in its routing table
```

Natolumin (approved):
A few notes I forgot earlier on the documentation, but otherwise LGTM

cmds/coredhcp/config.yml.example
skoef: I know of this option but I usually find it confusing. I updated the comments and mentioned the RFC.

Natolumin (approved):

PR #134, opened 2021-08-20
Added ability to add comments to static leases file in plugins/file

murphace:
Simple but useful change, add ability to have comments in static leases file.
Also added webhook plugin, can be used for dynamic dns updates or recordkeeping.


murphace:
@

murphace:
Will reissue pull request to a tag rather than master

PR #136, opened 2021-10-12
Add support for dependabot

gaby:
* Adds support for daily scans with dependabot.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


gaby:
@Natolumin It basically scans the gomod file and github-actions on a daily basis for outdated dependencies. It will auto create a PR with the update version.

Natolumin:
several of our dependencies are `v0.0.0-*` style. Would this give us a PR for every commit in the upstream repos ?

At the moment we don't really have a great setup for automated integration/e2e testing, so we've been updating dependencies when we do large changes and do more in-depth testing. So this will probably be interesting when the project is more mature about this kind of testing, but at the moment I think it would mostly create unactionable PRs that require large amounts of manual testing

I'll let the other maintainers chime in if they have more opinions on the subject, but for now I'd rather not go this way just yet

PR #137, opened 2021-10-12
support booting over HTTP(S) or FTP for DHCPv4

skoef:
In my environment where I boot using iPXE, the bootfile and further instructions are server over HTTP but the NBP plugin for CoreDHCP always assumes TFTP for DHCPv4. This MR changes that:

- when the argument passed to the NBP plugin is prefixed with http/https/ftp as parsed by `url.Parse`, it doesn't split the host/path into TFTPServerOption and BootfileOption, but passes the entire URL into BootfileOption
- for any other arguments the behaviour is identical as before


Natolumin:
I forgot to click the button...

PR #139, opened 2021-10-13
refresh records when leases file is updated

skoef:
The leases in my environment are fetched over HTTP from an API and are updated periodically. Since I don't want to restart CoreDHCP every time the leases change, this PR should make the file plugin automatically update the lease database when the file was changed. It does so in a best-effort: if it can't update the database it will show a warning but nothing will break.


Natolumin:
sorry for the long wait, this looks all good :)

skoef:
Thanks @Natolumin !

PR #144, opened 2021-10-26
fixed errors messages and added tests for netmask plugin

skoef:
While I was testing my coredhcp setup, I misconfigured the netmask plugin which made CoreDHCP panic. I spotted the bug in formatting the error messages and decided to write some tests while at it.


skoef:
@Natolumin could you review this?

PR #149, opened 2022-01-03
plugins: add mtu

Adphi:


cmds/coredhcp/main.go
```diff
@@ -20,6 +20,7 @@ import (
 	pl_dns "github.com/coredhcp/coredhcp/plugins/dns"
 	pl_file "github.com/coredhcp/coredhcp/plugins/file"
 	pl_leasetime "github.com/coredhcp/coredhcp/plugins/leasetime"
+	pl_mtu "github.com/coredhcp/coredhcp/plugins/mtu"
```
Natolumin, line 23: These need to be kept in sync between `cmds/coredhcp-generator` and `cmds/coredhcp` (this is the cause of the build failures)

plugins/mtu/plugin.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/coredhcp/coredhcp/plugins"
+)
+
+var log = logger.GetLogger("plugins/mtu")
+
+// Plugin wraps the DNS plugin information.
+var Plugin = plugins.Plugin{
+	Name:   "mtu",
+	Setup4: setup4,
+}
```
Natolumin, line 25: a small copy-paste mistake. I'd also like to add a comment mentioning that this is not relevant for dhcpv6 (to distinguish from plugins where we just haven't implemented it yet)
```suggestion
// Plugin wraps the MTU plugin information.
var Plugin = plugins.Plugin{
	Name:   "mtu",
	Setup4: setup4,
	// No Setup6 since DHCPv6 does not have MTU-related options
}
```

Natolumin (commented):
Hey there! thanks for the PR

I've put some minor comments inline, but in general this pr is good and I'll accept and merge it as soon as they're fixed
 
I can let you update to fix it if you want, or I can push the changes myself if you prefer but you'll need to enable the "[allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests)" option for that

cmds/coredhcp/main.go
Natolumin: ah, already fixed, thanks!

Adphi: Hi! yes, I missed that one 😉

plugins/mtu/plugin.go
Adphi: it's done.

PR #152, opened 2022-03-04
Allow two new features in IPv4 leases.txt: 1) Match on RFC3993 Subscriber-ID, 2) supply optional netmask,gateway alongside each IPv4

matthewbloch:
This is some work I did for a client who wanted coredhcp to do just a little bit more on their network. I hope the 2 commits are self-explanatory and well-tested enough.

I guess adding much more to the 'file' plugin might start to get messy, and need a proper parser, but right now it still seemed like overkill.


Natolumin:
Hey! Thanks for the contribution, but I don't think this is the direction we should take for the `file` plugin specifically. I have a few more thoughts in a comment on another issue: https://github.com/coredhcp/coredhcp/pull/147#issuecomment-1003706360 but specifically:

* I would rather not add fields to the plaintext line-based file format, especially as they are positional, and as you mentioned it gets very messy very quickly. The `Subscriber-ID` addition I think is fine as it doesn't restrict future syntax that much especially with the prefix, but the gateway/netmask definitely needs a real file format
* We won't add new IPv4 features without IPv6 equivalent unless the IPv6 equivalent doesn't make sense

I understand you may not want to make these changes yourself since this was paid work - in that case we'll probably cherry-pick the subscriber-id part if/when someone gets the equivalent ipv6 feature done

PR #171, opened 2023-10-13
[plugins/range] use sqlite3 instead of text file

insomniacslk:
The range plugin appends every new entry to a text file, without deduplication, and collisions are possible.
Implementing deduplication on a text file is either inefficient or tricky. Detecting collisions is inefficient, no matter if it's done in-memory and then rewriting the entire text file every time, or by parsing the text file every time.

Sqlite3 offers consistency, uniqueness without performance hit, and can enforce more complex constraints.

However using sqlite3 requires using CGo, which has a bunch of issues:
* it complicates cross-compiling
* builds are slower
* deployments are more complicated (no more single binary, not easily at least)

I have been running CoreDHCP with the sqlite3-based range plugin in my home network without issues so far, and no more duplication, nor collisions

### DIFF
--- a/plugins/range/storage.go
+++ b/plugins/range/storage.go
@@ -5,86 +5,87 @@
 package rangeplugin
 
 import (
-	"bufio"
+	"database/sql"
 	"errors"
 	"fmt"
-	"io"
 	"net"
-	"os"
-	"strings"
-	"time"
+
+	_ "github.com/mattn/go-sqlite3"
 )
+
+func loadDB(path string) (*sql.DB, error) {
+	db, err := sql.Open("sqlite3", fmt.Sprintf("file:%s", path))
+	if err != nil {
+		return nil, fmt.Errorf("failed to open database (%T): %w", err, err)
+	}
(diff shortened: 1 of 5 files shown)


plugins/range/storage.go
```diff
@@ -5,86 +5,88 @@
...
+
+	_ "github.com/mattn/go-sqlite3"
 )
 
+func loadDB(path string) (*sql.DB, error) {
+	db, err := sql.Open("sqlite3", fmt.Sprintf("file:%s", path))
+	if err != nil {
+		return nil, fmt.Errorf("failed to open database (%T): %w", err, err)
+	}
+	if _, err := db.Exec("create table if not exists leases4 (mac string not null, ip string not null, expiry int, primary key (mac, ip))"); err != nil {
```
insomniacslk, line 22: the table `lease4` is not configurable, but it might become if we see a use for it

insomniacslk:
I'll update the unit tests soon

pmazzini (approved):

PR #173, opened 2023-10-16
[plugins/range] fix SQL query for record persistence

insomniacslk:
The sql query to save a record to the database did just an insert but no update. This patch fixes it.


insomniacslk:
The `range` plugin is not stable yet, please keep tracking `master` before I create a release tag

Natolumin (approved):

PR #174, opened 2023-10-17
Replace use of deprecated io/ioutil

insomniacslk:
All the calls have been converted to their new equivalent.

### DIFF
--- a/cmds/coredhcp-generator/README.md
+++ b/cmds/coredhcp-generator/README.md
@@ -12,7 +12,7 @@ You can use `coredhcp-generator` to generate a `main.go` that includes all the
 plugins you wish. Just use it as follows:
 
 ```
-$ ./coredhcp-generator -from core-plugins.txt
+$ ./coredhcp-generator --from core-plugins.txt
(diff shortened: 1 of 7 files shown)


Natolumin (approved):
Purely out of curiosity, what minimum version of go does this make us require? I haven't kept up with go releases in a while

insomniacslk:
> Purely out of curiosity, what minimum version of go does this make us require? I haven't kept up with go releases in a while

The deprecation [started with Go 1.16](https://github.com/go-critic/go-critic/issues/1019), and we currently [require at least Go 1.9](https://github.com/coredhcp/coredhcp/commit/3e32ddb5ac865043b9beb2fb4127ba7e01f1c711) that I updated two months ago because of security vulnerabilities

insomniacslk:
Speaking of vulnerabilities, I'm going to update the dependencies because of two recent ones https://github.com/coredhcp/coredhcp/security/dependabot/22 and https://github.com/coredhcp/coredhcp/security/dependabot/21

PR #179, opened 2024-02-20
plugins/range: fixed memory leak in sqlite3 usage

insomniacslk:
I introduced a memory leak in the previous commit in the `range` plugin, by not closing a sqlite3 statement in a hot code path. This commit fixes it.

The graph below shows the effect of the fix on memory usage. The first annotation is a manual restart when the memory pressure was too high. The second restart marks the application of the patch. The correct behaviour is shown after the second restart, where the RSS usage flatlines after restarting.

Screenshot 2024-02-20 at 20 57 06

### DIFF
--- a/plugins/range/storage.go
+++ b/plugins/range/storage.go
@@ -35,8 +35,8 @@ func loadRecords(db *sql.DB) (map[string]*Record, error) {
 	defer rows.Close()
 	var (
 		mac, ip, hostname string
-		expiry  int
-		records = make(map[string]*Record)
+		expiry            int
+		records           = make(map[string]*Record)
 	)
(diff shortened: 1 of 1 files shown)


plugins/range/storage.go
```diff
@@ -64,6 +64,7 @@ func (p *PluginState) saveIPAddress(mac net.HardwareAddr, record *Record) error
 	if err != nil {
 		return fmt.Errorf("statement preparation failed: %w", err)
 	}
+	defer stmt.Close()
```
insomniacslk, line 67: this is the actual fix

pmazzini (approved):

PR #180, opened 2024-02-28
Update CI actions

Natolumin:
I noticed we have a bunch of warnings and errors in actions, including: 
* The lint action doesn't run because the yaml is malformed
* Codecov doesn't work for integration tests (marked as "unusable" here: https://app.codecov.io/github/coredhcp/coredhcp/commit/12a2f4060a7af9cfb3475abd38c79951ae55670f)
* A whole lot of deprecation notices

Update all the actions to their latest versions to address that. For the integration tests coverage I took the opportunity to bump the go version and use the new integration coverage support from go 1.20

Some additional details in individual commit messages


Natolumin:
The coverage decrease seems spurious, afaict from the codecov UI the diff seems to be that it used to count comment lines and now doesn't
Considering the coverage upload code and go version have changed, it's probably not meaningful to compare coverage before and after this commit

pmazzini (approved):

Issue #168, opened 2023-05-27
What is the repository status?


jmauro:
Hello,

It seems that the repository doesn't have a lot of recent activity, is this on purpose or the maintainers stopped this project?

Kind regards,
JM

rothgar:
I looks inactive based on the pull requests, open issues, and commit history.

Natolumin:
Being realistic, I'd say you shouldn't expect any activity in this repo

I haven't had contact with the other maintainers for several years now (I haven't tried very hard either) and they haven't been active here. I don't think they're completely disinterested in this field considering the dhcp library is still actively maintained, but I can't speak for them obviously.

Although I think I would like to keep maintaining this repo, I do not have full access to it, only basic committer permissions - I can accept submitted PRs, but not push or accept my own code (which is very limiting) or otherwise change repo configuration.

There are people who seemed interested in taking up a fork in a different repo in a few recent PRs, and I'd say in the short term that's your best bet

insomniacslk:
I am planning to resume work on CoreDHCP now, after I got burned once again by other implementations. As a first step I'll try to stabilize and clean up the core plugins (where needed); then I want to resume the work on offering a better configuration language (either via https://github.com/coredhcp/coredhcp/issues/89 or https://github.com/coredhcp/coredhcp/issues/91 ). Any help from @Natolumin will be critical to do a better job, and highly appreciated, but I'm OK with working on this alone, hopefully with community input.

farcaller:
@insomniacslk I would appreciate having some release tags moving forward. Anything is better than just randomly tracking the master. I was about to add a nix package for this earlier today as I got burned by dnsmasq issues, and I learned there's not a single "release".

insomniacslk:
@farcaller I agreee with release tags. This was a side project and I didn't expect more than me and a few others to use it, but judging by forks and stars on this repo, it deserves proper love.

brennoo:
I've created a docker image [here](https://github.com/brennoo/docker-coredhcp) - testing it on my home lab at the moment.
as we don't have the first release yet I'm tagging the images with the commit hashes of this repo - I added a go.mod and a dependabot config aiming to update the image whenever there is a push here :-)

insomniacslk:
Thanks @brennoo for that! I found some other issues with expired leases in the `range` pluginm, hence no v1.0.0 release yet. Testing this kind of things takes a long time for issues to show up, but I guess once this one is cleared up we are good to release. CC @Natolumin

[...]

PR #212, opened 2024-12-15
Upload build artifacts

insomniacslk:
The build artifacts will show up in the artifacts section of the workflow, see https://github.com/actions/upload-artifact?tab=readme-ov-file#where-does-the-upload-go


insomniacslk:
Artifacts uploaded successfully, https://github.com/coredhcp/coredhcp/actions/runs/12338687023

Fixes #168

insomniacslk:
I can't tell why the lint step is failing - it works locally, and it doesn't print any error on the CI, but the task reports a failure. Going to merge anyway

PR #229, opened 2025-03-19
Fixed example: reference to sqlite3 db rather than old txt file

insomniacslk:
The example was using the extension ".txt" for what actually is a sqlite3 file


insomniacslk:
The linter failure is unrelated to the change and needs to be fixed separately

PR #230, opened 2025-03-20
support v4 dhcp inform messages

christophbrejla:
I need the possibility in a custom plugin to handle DHCP INFORM messages. The server/handle.go did not have such case. I think this could be useful also for other use cases and should be generally supported.


server/handle.go
```diff
@@ -127,6 +127,9 @@ func (l *listener4) HandleMsg4(buf []byte, oob *ipv4.ControlMessage, _peer net.A
 		tmp.UpdateOption(dhcpv4.OptMessageType(dhcpv4.MessageTypeOffer))
 	case dhcpv4.MessageTypeRequest:
 		tmp.UpdateOption(dhcpv4.OptMessageType(dhcpv4.MessageTypeAck))
+	case dhcpv4.MessageTypeInform:
+		tmp.UpdateOption(dhcpv4.OptMessageType(dhcpv4.MessageTypeAck))
+		tmp.YourIPAddr = nil
```
Natolumin, line 132: This seems superfluous, are you sure this is needed? It should already be nil at this point, and in any case other plugins that might override it run later. 
```suggestion
```

Natolumin (approved):
Thanks! This would be good to have.

I have one minor question inline but otherwise lgtm

christophbrejla:
Hello @Natolumin it has been a while, would be able to merge this one now? thanks

Natolumin (dismissed):
@christophbrejla Please read through https://developercertificate.org/ an add a `signed-off-by: ` line in your commit messages accordingly (eg with `git rebase master --signoff`) so we can merge

Natolumin:
Thank you for your contribution!

PR #253, opened 2025-08-23
plugin/file now checks for duplicates and allows trailing comments in its input

rickb777:
This PR increases the sanity checks in the external files that configure fixed IP address mappings and allows more flexible comment syntax.

1. MAC addresses: duplicates are highlighted as ~an error~ a warning
2. IP addresses: duplicates are highlighted as ~an error~ a warning
3. Trailing comments are now allowed.
4. The documentation has been augmented.

All duplicates are reported in the resulting ~error~ warning messages, if there are any.

Also note that the unit tests now explicitly close their test files before reading them, making it slightly more obvious that the file is not being modified when the functions under test execute.

### DIFF
--- a/plugins/file/plugin.go
+++ b/plugins/file/plugin.go
@@ -47,6 +47,7 @@ import (
 	"net"
 	"net/netip"
 	"os"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -138,22 +139,24 @@ func loadDHCPRecords(filename string, protVer int, check func(netip.Addr) bool) 
 		addresses[tokens[1]]++
 	}
 
-	duplicates := duplicatesAsErrors(addresses)
-	if len(duplicates) > 0 {
-		return nil, errors.Join(duplicates...)
-	}
+	duplicatesWarning(addresses)
 
 	return records, nil
 }
 
-func duplicatesAsErrors(ipAddresses map[string]int) []error {
-	var duplicates []error
+func duplicatesWarning(ipAddresses map[string]int) {
+	var duplicates []string
 	for ipAddress, count := range ipAddresses {
 		if count > 1 {
-			duplicates = append(duplicates, fmt.Errorf("address %s is in %d records", ipAddress, count))
+			duplicates = append(duplicates, fmt.Sprintf("Address %s is in %d records", ipAddress, count))
 		}
 	}
-	return duplicates
+
+	sort.Strings(duplicates)
+
+	for _, warning := range duplicates {
+		log.Warningf(warning)
+	}
 }
 
 // Handler6 handles DHCPv6 packets for the file plugin
--- a/plugins/file/plugin_test.go
+++ b/plugins/file/plugin_test.go
@@ -91,7 +91,7 @@ func TestLoadDHCPv4Records(t *testing.T) {
 		assert.Error(t, err)
 	})
 
-	t.Run("duplicate MAC address should raise error", func(t *testing.T) {
+	t.Run("duplicate MAC address are allowed", func(t *testing.T) {
 		// setup temp leases file
 		tmp, err := os.CreateTemp("", "test_plugin_file")
 		require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


Natolumin (commented):
Thank you for your contribution!

I appreciate most of the refactors and improvements in here, and would gladly merge them independently; especially trailing comments and the use of the newer netip types

The main functional change of rejecting duplicates however I don't believe is desirable at this point:
* It represents a breaking change in the file format, rejecting previously accepted lease files. That's certainly not a dealbreaker on its own but I'm not convinced there's a compelling reason to do this
* Rejecting duplicate IPs removes functionality where it was previously possible to assign the same IP to multiple MACs. dnsmasq for example explicitly supports this (with the expected caveat that that assumes both MACs do not coexist on the network and it will cause issues if they do)
* Rejecting duplicate MACs removes an otherwise useful functionality where updated mappings can just be appended to the file (potentially with an automated script) without having to independently parse and garbage collect the mapping, you just had the last entry "win"

Not opposed to doing it anyway still (I think those are fairly weak reasons), but I don't see a compelling reason to reject the duplicates. What was your motivation for the change initially?

plugins/file/plugin.go
```diff
@@ -4,35 +4,50 @@
...
+//
+//	$ cat leases_v6.txt
+//	# IPv6 fixed addresses
+//	00:11:22:33:44:55 2001:db8::10:1
+//	A1:B2:C3:D4:E5:F6 2001:db8::10:2  # uppercase is only permitted for MAC
+//
+// Any text following '#' is a comment that is ignored. MAC addresses can be upper or lower case.
+// IPv6 addresses must use lowercase, as per RFC-5952.
+//
+// Each MAC or IP address must be unique within the file.
```
Natolumin: outdated comment

plugins/file/plugin.go
```diff
@@ -4,35 +4,50 @@
...
+//	   plugins:
+//	     - file: "file_leases.txt" [autorefresh]
+//	   ...
 //
 // If the file path is not absolute, it is relative to the cwd where coredhcp is run.
 //
 // Optionally, when the 'autorefresh' argument is given, the plugin will try to refresh
 // the lease mapping during runtime whenever the lease file is updated.
+//
+// Note that the file plugin must come after any general plugins needed, e.g. dns or router.
```
Natolumin: confusingly, this is only true for DHCPv4 (this stops processing in DHCPv4 because you can only return a single IP, but DHCPv6 can have later plugins add other `IA_` options)

plugins/file/plugin.go
```diff
@@ -4,35 +4,50 @@
...
-//  01:23:45:67:89:01 10.0.10.10
+//	$ cat leases_v4.txt
+//	# IPv4 fixed addresses
+//	00:11:22:33:44:55 10.0.0.1
+//	a1:b2:c3:d4:e5:f6 10.0.10.10  # lowercase is permitted
+//
+//	$ cat leases_v6.txt
+//	# IPv6 fixed addresses
+//	00:11:22:33:44:55 2001:db8::10:1
+//	A1:B2:C3:D4:E5:F6 2001:db8::10:2  # uppercase is only permitted for MAC
```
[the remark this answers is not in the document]
rickb777: I deleted the trailing comment

plugins/file/plugin.go
```diff
@@ -156,26 +174,26 @@ func Handler6(req, resp dhcpv6.DHCPv6) (dhcpv6.DHCPv6, bool) {
...
 
+	// note that net.HardwareAddr.String() uses lowercase hexadecimal
 	ipaddr, ok := StaticRecords[mac.String()]
 	if !ok {
-		log.Warningf("MAC address %s is unknown", mac.String())
+		log.Infof("MAC address %s is unknown", mac)
 		return resp, false
 	}
-	log.Debugf("found IP address %s for MAC %s", ipaddr, mac.String())
+	log.Printf("MAC address %s given IP address %s", mac, ipaddr)
```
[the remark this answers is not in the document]
rickb777: All Printf cases have been changed to Infof.

plugins/file/plugin.go
rickb777: Changed to

Each MAC or IP address should normally be unique within the file. Warnings will be logged for any duplicates.

